PR #43, opened 2019-02-18
add format specifier support

goatshriek:
Add support for format specifiers in log messages, entries, and windows event log insertion strings.


goatshriek:
Diff coverage is expected to be low due to inability of codecov to take Windows test coverage into account.

PR #100, opened 2020-07-31
improve documentation for new developers

goatshriek:
There are several gaps in the documentation that do not cover situations that someone new to the project is likely to encounter. This set of changes updates the documentation so that it is easier for those looking to contribute, perhaps by solving a "good first issue", have an easier time getting started.


pedestrianlove (commented):
Details are mostly perfect, just that my 2 cents:
1. It would be great if I can see a link to `docs/CONTRIBUTING.md` when I first entered this repo(in `README.md`).
If I don't fiddle with the `docs` folder in the project, I will never know the existence of this file.
I initially guessed that it was for the documentation of the library usage itself!

2. It would be even better if `docs/development.md` (or link to somewhere else) can contain some simple instructions on how to use `gtest`. ( as simple as the building instruction in `INSTALL.md` if possible)
I found out how to somewhat run `gtest` in this project only after I went through this route:
TRAVIS-CI's config file -> `linux_script.sh` -> `make check` -> check in the `Makefile` that the `check` build target is the job that runs `gtest` -> `make check == running gtest in this project` confirmed.
(Since I missed the `docs/CONTRIBUTING.md`, I also missed `development.md`)

Finally, it is absolutely nice to NOT put all the content in the `README.md`, since most of the visitors of this would be looking for normal usage. Just that, open a little door to the `CONTRIBUTING.md` would be pretty niche.

I can't think of other obvious weak spots for welcoming newb developers like me.

pedestrianlove:
Just another thought...
For a normal user that just want to use the library, I would suggest that you can put maybe 1 or 2 most specialized usages in the `README.md`, and put the remaining examples in something like `basic_examples.md`, `advanced_examples.md`.

IMO, since `README.md` is the first page that the newcomer will see, the main works that it should do should be:
1. Direct their flow of browsing
i.e. Point them to other requested subfiles in the repo, like `CONTRIBUTING.md`, ``

2. Provide some simple info if possible.
e.g. provide them some quick info like build status, simple build instructions etc.

3. Provide densified info on why they should be interested in this library.
It shouldn't be too long or they may just close this tab or go to the previous page.

You can also try to think of your own way to form your own style of `README.md`, something like this(from top to down):
1. Abstract about this library.
2. How to install(maybe 3-4 lines of command).
3. Usage examples remain the same as the current README.md.
4. More info on contributing(links to `docs/CONTRIBUTING.md`).

docs/CONTRIBUTING.md
```diff
@@ -32,17 +32,20 @@ these guidelines:
    of `git commit`).
  * Keep the number of commits in your request to a minimum. Having multiple
    commits is fine if necessary to separate multiple changes, but make sure
-   that these are logical and value to the commit history.
+   that these are logical and add value to the commit history. For smaller
+   changes, your request may simply be squash merged, making sure to preserve
```
pedestrianlove, line 36: You can also link to [the helpful guide](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) I used when I was trying to reduce my garbage commit (something like fixing a typo etc).

Aside from this one, I do think this PR is good to go for me. :)

pedestrianlove (approved):

docs/CONTRIBUTING.md
goatshriek: Great, thank you very much for reviewing this! I will merge it after I have time to do another full review on it myself, and update the issues I have marked as "help wanted" to link to the Contribution and development guides as starting points.

PR #101, opened 2020-08-02
Added stumpless_version_cmp

### LINKED ISSUE #98: comparison function for versions
While the library provides a way to get the current version as a string or a custom struct, it does not provide an easy way to compare versions using semantic versioning principles. This functionality would be convenient for any users wishing to put a version guard in their code, for example to throw warnings if a major version number is incremented unintentionally. This would be a C-like version of the version specifications provided in Gemfiles in Ruby or requirements.txt files in Python.

This function should be called something like `stumpless_version_cmp` and should take two const `stumpless_version` structs as arguments. The return result will be an int following the same conventions as `strcmp` regarding less than, equal, or greater than. Only the major, minor, and patch values need to be compared - do not worry about semantic versioning concepts like build metadata or pre-release versions.

This is a good first issue, especially for anyone looking to get a stronger understanding of [Semantic Versioning](https://semver.org/).

Check out the `include/stumpless/version.h` header and `src/version.c` source module to get a feel for the version struct and what an implementation of this might look like. The version test suite already exists at `test/function/version.cpp` and would just need to have a few tests added to it for the new function.

pedestrianlove:
This should fix #98 .

1. Added `stumpless_version_cmp`
 And its docs.

2. Added entry in `src/windows/stumpless.def`

3. Added test for `stumpless_version_cmp`
Null version case and basic test case.

### DIFF
--- a/include/stumpless/version.h
+++ b/include/stumpless/version.h
@@ -45,6 +45,21 @@ struct stumpless_version *
 stumpless_get_version( void );
 
 /**
+ * Compares the version of two given stumpless_version struct.
+ * @param version_x,version_y Two given version(const) to be compared.
+ * `
+ * @return 0 if version numbers are the same, <br />
+ *         1 if version_y->patch > version_x->patch,   <br />
+ *         10 if version_y->minor > version_x->minor,  <br />
+ *         100 if version_y->major > version_x->major, <br />
+ *         666 if one of the version pointers is null. <br />
+ */
+int
+stumpless_version_cmp( const struct stumpless_version * version_x, 
+                        const struct stumpless_version * version_y);
+
+
+/**
  * Creates a string representation of the given version.
  *
  * This string will be NULL-terminated, and will appear in the standard
--- a/include/stumpless/version.h
+++ b/include/stumpless/version.h
@@ -45,18 +45,21 @@ struct stumpless_version *
 stumpless_get_version( void );
 
 /**
- * Compares the version of two given stumpless_version struct.
+ * Compares the version of two given stumpless_version struct. 
+ * 
+ * This function make use of a similar convention used by the standard library's strcmp function.
+ * 
  * @param version_x,version_y Two given version(const) to be compared.
- * `
- * @return 0 if version numbers are the same, <br />
- *         1 if version_y->patch > version_x->patch,   <br />
- *         10 if version_y->minor > version_x->minor,  <br />
- *         100 if version_y->major > version_x->major, <br />
- *         666 if one of the version pointers is null. <br />
+ * 
+ * @return 0 if version numbers are the same,                           <br />
+ *         1 if version_x->patch - version_y->patch > 0,    -1   if < 0 <br />
+ *         10 if version_x->minor - version_y->minor > 0,   -10  if < 0 <br />
+ *         100 if version_x->major - version_y->major > 0,  -100 if < 0 <br />
+ *         INT_MAX if one of the version pointers is null.              <br />
  */
 int
 stumpless_version_cmp( const struct stumpless_version * version_x, 
-                        const struct stumpless_version * version_y);
+                       const struct stumpless_version * version_y );
 
 
 /**
--- a/src/version.c
+++ b/src/version.c
@@ -41,6 +41,31 @@ stumpless_get_version( void ) {
   return version;
 }
 
+int
+stumpless_version_cmp ( const struct stumpless_version * version_x, 
+                         const struct stumpless_version * version_y) {
+  if ( version_x == NULL || version_y == NULL ) {
+    return 666;
+  }
+  
+  if( version_y->major != version_x->major ) {
+    return ( version_y->major - version_x->major )/
+	    ( version_y->major - version_x->major ) * 100;
+  }
+  
+  if( version_y->minor != version_x->minor ) {
+    return ( version_y->minor - version_x->minor )/
+	    ( version_y->minor - version_x->minor ) * 10;
+  }
+  
+  if( version_y->patch != version_x->patch ) {
+    return ( version_y->patch - version_x->patch )/
+	    ( version_y->patch - version_x->patch ) * 1;
+  }
+
+  return 0;
+}
+
 char *
 stumpless_version_to_string( const struct stumpless_version *version ) {
   struct strbuilder *builder;
--- a/src/version.c
+++ b/src/version.c
@@ -17,7 +17,6 @@
  */
 
 #include <stddef.h>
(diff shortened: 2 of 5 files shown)


pedestrianlove:
I just fixed my excess commit from the previous PR.
This should work neatly now.

include/stumpless/version.h
```diff
@@ -44,6 +44,21 @@ struct stumpless_version {
 struct stumpless_version *
 stumpless_get_version( void );
 
+/**
+ * Compares the version of two given stumpless_version struct.
+ * @param version_x,version_y Two given version(const) to be compared.
+ * `
```
goatshriek: looks like there is a stray backtick character here that can be removed

include/stumpless/version.h
```diff
@@ -44,6 +44,21 @@ struct stumpless_version {
...
+ * @return 0 if version numbers are the same, <br />
+ *         1 if version_y->patch > version_x->patch,   <br />
+ *         10 if version_y->minor > version_x->minor,  <br />
+ *         100 if version_y->major > version_x->major, <br />
```
goatshriek: Be sure to update this documentation with info about positive/negative return values. You could do this with one sentence describing the positive/negative return meaning, and a second with the 1/10/100 meaning. Or some other way that seems good to you, as long as it's clear that both positive and negative values are possible.

include/stumpless/version.h
```diff
@@ -44,6 +44,21 @@ struct stumpless_version {
...
+ *         666 if one of the version pointers is null. <br />
```
goatshriek: I like the humor in the choice of `666` here, but I think a more logical choice would be `INT_MIN` or `INT_MAX`.

include/stumpless/version.h
```diff
@@ -44,6 +44,21 @@ struct stumpless_version {
...
+ */
+int
+stumpless_version_cmp( const struct stumpless_version * version_x, 
+                        const struct stumpless_version * version_y);
```
goatshriek: remove the extra space character here to make sure the lines align

src/version.c
```diff
@@ -41,6 +41,31 @@ stumpless_get_version( void ) {
...
+  }
+  
+  if( version_y->minor != version_x->minor ) {
+    return ( version_y->minor - version_x->minor )/
+	    ( version_y->minor - version_x->minor ) * 10;
+  }
+  
+  if( version_y->patch != version_x->patch ) {
+    return ( version_y->patch - version_x->patch )/
+	    ( version_y->patch - version_x->patch ) * 1;
```
goatshriek: I suspect that you can remove the multiplication by 1

test/function/version.cpp
```diff
@@ -87,6 +87,41 @@ namespace {
...
+    struct stumpless_version version_y = {1, 5, 1};
+
+    EXPECT_EQ( stumpless_version_cmp( &version_x, &version_y ), 1 );
+  }
+
+  TEST( VersionCmp, BasicMinor ) {
+    struct stumpless_version version_x = {1, 5, 0};
+    struct stumpless_version version_y = {1, 6, 0};
+
+    EXPECT_EQ( stumpless_version_cmp( &version_x, &version_y ), 10 );
```
goatshriek: this should return -10 instead (because the first arg is smaller)

test/function/version.cpp
```diff
@@ -87,6 +87,41 @@ namespace {
...
+  }
+
+  TEST( VersionCmp, BasicMajor ) {
+    struct stumpless_version version_x = {1, 5, 0};
+    struct stumpless_version version_y = {2, 6, 0};
+
+    EXPECT_EQ( stumpless_version_cmp( &version_x, &version_y ), 100 );
```
goatshriek: this should return -100 instead (because the first arg is smaller)

test/function/version.cpp
```diff
@@ -87,6 +87,41 @@ namespace {
...
+    struct stumpless_version version_y = {1, 5, 0};
+
+    EXPECT_EQ( stumpless_version_cmp( &version_x, &version_y ), 0 );
+  }
+
+  TEST( VersionCmp, BasicPatch ) {
+    struct stumpless_version version_x = {1, 5, 2};
+    struct stumpless_version version_y = {1, 5, 1};
+
+    EXPECT_EQ( stumpless_version_cmp( &version_x, &version_y ), 1 );
```
goatshriek, line 102: positive 1 is fine here, because the first arg is larger

include/stumpless/version.h
```diff
@@ -44,6 +44,21 @@ struct stumpless_version {
 struct stumpless_version *
 stumpless_get_version( void );
 
+/**
+ * Compares the version of two given stumpless_version struct.
```
goatshriek: Make sure that the function returns a positive _or_ negative value depending on which of the two arguments is larger: negative if the first is smaller, positive if the first is larger. This is the same convention used by the standard library's `strcmp` function (some good documentation on it [here](http://www.cplusplus.com/reference/cstring/strcmp/)), and so should be familiar to users.

This behavior is important because it allows the caller to know which of the two versions is greater, so that they can make decisions such as "it's fine if these aren't equal as long as the actual version is higher than what I expected".

src/version.c
```diff
@@ -41,6 +41,31 @@ stumpless_get_version( void ) {
   return version;
 }
 
+int
+stumpless_version_cmp ( const struct stumpless_version * version_x, 
+                         const struct stumpless_version * version_y) {
+  if ( version_x == NULL || version_y == NULL ) {
+    return 666;
```
goatshriek: before returning the error code, be sure to call `raise_argument_empty` to set the library error code appropriately first. 

I apologize once again: this is something I haven't documented well and I'll add a section to the development guide about it. Until I get some time to do that, check out the `src/param.c` file for lots of examples for doing this to signify that an argument was empty. The short version is that all functions (except for destructors and error handling code itself) raise errors using the `raise_` family of internal functions in `include/private/error.h` when they need to.

goatshriek (changes requested):
Excellent work! I especially like the use of different magnitudes to signify which part of the version was different - it isn't something I had thought of but is definitely a good feature!

Most of the following comments are minor edits, with the exception of making sure that this function follows the same positive or negative return value convention that strcmp does. Even that shouldn't be much of a change.

src/version.c
pedestrianlove: Thanks for your detailed reply! I'll get back to you after I've done all these changes.

goatshriek (approved):

goatshriek:
Looks good, thanks again for all of your work on this!

PR #110, opened 2020-08-30
implementing STUMPLESS_OPTION_PID

### LINKED ISSUE #77: implement PID option
Stumpless currently does not support the `STUMPLESS_OPTION_PID` option despite providing it in its header files. This could easily be implemented by checking the target's setting for this option during formatting, and using `NILVALUE` (a dash character) if this option is not set.

This option should be off by default, being enabled by `stumpless_set_option` on a target. This is a change in current behavior, but brings the library closer to the behavior described by [`syslog`](https://man7.org/linux/man-pages/man3/syslog.3.html), which this library currently attempts to stay in line with where possible and reasonable.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here and someone can help you get past the stumbling block. You are also free to take a different approach if you feel it is better - this is merely meant as a starting point.

First, have a look at the [documentation](https://goatshriek.github.io/stumpless/docs/c/latest/target_8h.html) for `stumpless_get_option`, `stumpless_set_option`, and `stumpless_unset_option` in `include/stumpless/target.h`. You will not need to modify any of these, but you will use them in your implementation so they are important to understand.

Next, check out `stumpless_add_entry` in `src/target.c`. This is the workhorse function for logging messages, and includes the calls to format messages before sending them. The formatting call currently only uses the `stumpless_entry` struct to create the message, but this will not be enough to tell if the target's option is set.

There are a few possible solutions. You could add a completely new formatting function (something like `format_entry_without_pid`) to `src/formatter.c` that does not use the procid of the entry. You could change the current formatting function to also take a pointer to the `stumpless_target` struct that it can use to check the option while formatting. You could change the current formatting function to simply take a boolean switch on whether to use the pid or not. Or you could solve it  a different way. The specific approach is up to you - just make sure that you can explain why you feel it's the best approach. The end result should be that if `STUMPLESS_OPTION_PID` is not set on the target, then the `PROCID` field of the logged message is a single dash character, the RFC 5424 NILVALUE.

Once this is done, add some tests to make sure that the behavior is as intended. Test code already exists to make sure that setting and unsetting the option behaves as expected - look for `TEST( SetOption, Pid )` in `test/function/target.cpp`. Use this test as a starting point to create a new test that logs a message before and after setting the option. The message logged before the option is set should not include the pid. The message logged after should have it. To detect whether the pid section is a dash or a value will require some work - you could write your own regex, or check out the [Google Mock matchers](https://github.com/google/googletest/blob/master/googlemock/docs/cheat_sheet.md) to perhaps make this easier. For details on the format of the messages, you can refer to the [RFC](https://tools.ietf.org/html/rfc5424#section-6) that stumpless follows (don't let the RFC intimidate you, the format section is pretty easy to follow).

TylerStanish:
This should implement STUMPLESS_OPTION_PID. By default this is disabled (will not display the PID). Should fix #77.

### DIFF
--- a/include/private/formatter.h
+++ b/include/private/formatter.h
@@ -20,6 +20,7 @@
 #  define __STUMPLESS_PRIVATE_FORMATTER_H
 
 #  include <stumpless/entry.h>
+#  include <stumpless/target.h>
 #  include "private/strbuilder.h"
 
 #  define RFC_5424_FULL_DATE_BUFFER_SIZE 11
@@ -32,7 +33,10 @@
 #  define RFC_5424_TIMESTAMP_BUFFER_SIZE 33
 #  define RFC_5424_WHOLE_TIME_BUFFER_SIZE 20
 
+#  define RFC_5424_NILVALUE '-'
+
 struct strbuilder *
-format_entry( const struct stumpless_entry *entry );
+format_entry( const struct stumpless_entry *entry,
+              struct stumpless_target *target );
 
 #endif /* __STUMPLESS_PRIVATE_FORMATTER_H */
--- a/include/stumpless/option.h
+++ b/include/stumpless/option.h
@@ -30,7 +30,7 @@
 
 #    include <syslog.h>
 
-/** Not currently supported. PIDs are always included in stumpless messages. */
+/** Option to include the PID in stumpless messages  */
 #    define STUMPLESS_OPTION_PID    LOG_PID
 /** Not currently supported. */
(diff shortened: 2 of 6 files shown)


goatshriek:
Excellent work! Only one thing needs to change before this is accepted - please update the documentation for the PID option in `include/stumpless/option.h` to reflect the implementation with a brief summary of what the option does. Currently it is marked as "not supported" which will be incorrect after this. I'll then rebuild the doxygen documentation once this is merged.

Issue #113, opened 2020-09-02
length checking of the msgid


goatshriek:
The syslog RFC ([RFC 5424](https://tools.ietf.org/html/rfc5424)) specifies a given length for many fields of a log message, including the msgid. Checking the length of supplied msgid arguments will not only help users stay compliant with the standard, but will also allow for efficiencies in the implementation by eventually allowing for the use of fixed-length buffers instead of memory allocated on the heap.

As specified in the RFC, msgid parameters should be checked for their length and rejected if they are more than the allowed length (32 characters). Note that this length does not include the terminating NULL character from the C string representation.

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block. You are also free to take a different approach if you feel it is better - this is merely meant as a starting point.

There are a number of functions that allow a msgid to be supplied - the constructor for entries, the set feature for the msgid of entries, and the default msgid setting for targets, to list a few. Since this check will need to take place in several places, you will be best served by implementing it in a single function and then calling it wherever required.

Declare your new function (something like `validate_msgid_length`) in `include/private/validate.h`, and define it in `src/validate.c`. You will need to create the latter of these files - don't forget to add it to the source file list `STUMPLESS_SOURCES` in `CMakeLists.txt`. After it has been defined, add calls to it wherever a msgid is passed as a parameter in the library, which are the following functions:
 * `vstumpless_new_entry` (in `src/entry.c`)
 * `stumpless_set_entry_msgid` (in `src/entry.c`)
 * `stumpless_set_target_default_msgid` (in `src/target.c`)

An error already exists that should be used for invalid message ids: `STUMPLESS_ARGUMENT_TOO_BIG`. Raise this error in your validation function if a msgid that is too long is detected, using the `raise_argument_too_big` function from `src/error.c`.

Don't forget to add tests for your new functionality! These should go alongside the tests for the functions that have msgids - check `test/function/target.cpp` and `test/function/entry.cpp` for existing tests for these functions, and add new tests that make sure that msgids that are too long are rejected.

LambdaApproach:
[...]

PR #125, opened 2020-10-12
Msgid length check functionality

LambdaApproach:
This pull request is intended to resolve issue #113 

General overview of what's been done:

1. Created a validate_msgid_length function to prevent msgids exceeding the 32 character limit from being accepted. You can see this in validate.h and the newly created validate.c

2. Included error handling in case msgids are rejected within the newly created function, and created a new error message and code type to indicate the type of error clearly.

3. Added 2 tests in entry.cpp and 1 test in target.cpp to check that the new function is working as intended.

Hope this is good, if not I'll try to fix whatever is wrong in a reasonable time frame.


LambdaApproach:
I checked some of those Travis CI results and it looks like I got overzealous with my include directives in validate.c, I'll work on that and also find a way to fix the signed commit issue as well.

goatshriek:
No worries! You can also leave your pull requests open and simply push more commits to them as you fix things if that is easier for you instead of opening a fresh one - either way works. This is a small enough change that it will get squashed into a single commit with you as the author, so don't worry about the commit history or anything like that.

LambdaApproach:
Got it, thanks. I fixed the commit sign-off issue and I think I got the include directives cleaned up as well. I did force push in order in order to amend my commits so it won't let me reopen this pull request.

I'll open a new one and keep it open even if any more issues come up later today. Apologies for the hassle.

PR #126, opened 2020-10-12
Length check for msgid

LambdaApproach:
Second attempt after the premature closure of the other pull request. Commits have been signed off and includes have been cleaned up so hopefully this one goes a little better.

This is intended to resolve issue #113

General overview of what's been done:

1. Created a validate_msgid_length function to prevent msgids exceeding the 32 character limit from being accepted. You can see this in validate.h and the newly created validate.c

2. Included error handling in case msgids are rejected within the newly created function, and created a new error message and code type to indicate the type of error clearly.

3. Added 2 tests in entry.cpp and 1 test in target.cpp to check that the new function is working as intended.


include/private/config/locale/en-us.h
```diff
@@ -127,12 +127,18 @@ ARG_NAME " was NULL"
 "checked to see if an unsupported target type was open"
 
 #  define L10N_WINDOWS_SOCKET_ERROR_CODE_TYPE \
-"Windows Socket error code"
+"Windows Socket error code"
```
goatshriek: it looks like you've unintentionally added some trailing whitespace here - be sure to trim it off

include/private/config/locale/en-us.h
```diff
@@ -127,12 +127,18 @@ ARG_NAME " was NULL"
...
-"Windows Socket error code"
+"Windows Socket error code" 
 
 #  define L10N_WINSOCK2_SOCKET_FAILED_ERROR_MESSAGE \
 "failed to open a winsock2 socket"
 
 #  define L10N_WSAGETLASTERROR_ERROR_CODE_TYPE \
 "the result of WSAGetLastError after the failed call"
 
+#  define L10N_STRING_TOO_LONG \
```
goatshriek: I know this will be a bit of work to change in every locale and usage in the source code, so I'll deem this optional, but consider making the name of this symbol `L10N_STRING_TOO_LONG_ERROR_MESSAGE` so that it is consistent with all of the others.

src/entry.c
```diff
@@ -486,6 +486,10 @@ stumpless_set_entry_msgid( struct stumpless_entry *entry,
   VALIDATE_ARG_NOT_NULL( entry );
 
   effective_msgid = msgid ? msgid : "-";
+  if( !validate_msgid_length( effective_msgid ) ) {
+      return false;
```
goatshriek: While `false` will likely be the same as `NULL` due to the requirement in POSIX that `NULL` evaluates to `false` in a conditional, it would be more portable (and consistent with the surrounding code) to return `NULL` in this error condition instead of `false`.

src/entry.c
```diff
@@ -690,6 +694,10 @@ vstumpless_new_entry( int facility,
   }
 
   effective_msgid = msgid ? msgid : "-";
+  if( !validate_msgid_length( effective_msgid ) ) {
+      goto rejected_msgid;
```
goatshriek: Use the existing `fail_msgid` label here instead of adding a new one. As is, there is a memory leak where the `entry` and `app_name` will be allocated, but not freed upon an error condition. The existing failure branch already handles this case for you.

src/target.c
```diff
@@ -347,6 +347,10 @@ stumpless_set_target_default_msgid( struct stumpless_target *target,
   VALIDATE_ARG_NOT_NULL( target );
   VALIDATE_ARG_NOT_NULL( msgid );
 
+  if( !validate_msgid_length( msgid ) ) {
+      return false;
```
goatshriek: return `NULL` here for the same reason as in `stumpless_set_entry_msgid`

goatshriek (changes requested):
Overall, excellent work! There are a few tweaks that I've outlined here that I'd like you to make before I merge this into the default branch.

goatshriek (approved):

LambdaApproach:
I've added the proper security keys to enable proper signing of commits. I think I found a way to amend all the previous commits to be signed but I can't find a way to push it unless I force it. Should I proceed in doing so, or would you rather I try something else?

goatshriek:
You don't need to worry about signing the commits with GPG or anything like that, they will simply be squashed anyway. The "signed-off-by" tag requirement is a holdover from older times when Github authorship of commits wasn't handled the same as it is now and was the only way to give you credit - perhaps I should remove that note in the documentation now. You'll be listed as the author of the resulting commit, which is the intention of that requirement: to make sure you get contributor credit.

I'll merge this in later today once the CI builds are completed.

Thanks so much for your work on adding this feature and your (extremely) timely responses!

Issue #114, opened 2020-09-03
format checking of msgid


goatshriek:
The syslog RFC ([RFC 5424](https://tools.ietf.org/html/rfc5424)) specifies that a number of fields only include printable ASCII characters (decimal values 33 to 126 inclusive), including the msgid. Checking the characters of supplied msgid arguments will not only help users stay compliant with the standard and provide one point of mitigation for encoding bugs/vulnerabilities.

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block. You are also free to take a different approach if you feel it is better - this is merely meant as a starting point.

There are a number of functions that allow a msgid to be supplied - the constructor for entries, the set feature for the msgid of entries, and the default msgid setting for targets, to list a few. Since this check will need to take place in several places, you will be best served by implementing it in a single function and then calling it wherever required.

Declare your new function in `include/private/entry.h`, and define it in `src/entry.c`. After it has been defined, add calls to it wherever a msgid is passed as a parameter in the library, which are the following functions:
 * `vstumpless_new_entry` (in `src/entry.c`)
 * `stumpless_set_entry_msgid` (in `src/entry.c`)
 * `stumpless_set_target_default_msgid` (in `src/target.c`)

You will need to add a new error code that can be thrown when an issue is detected. You can find information on the error codes and handling in the appropriate [section](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#error-handling) of the development guide. Name the new error something like 'STUMPLESS_INVALID_ENCODING' so that it can be reused by encoding checks of other values in the future.

Don't forget to add tests for your new functionality! These should go alongside the tests for the functions that have msgids - check `test/function/target.cpp` and `test/function/entry.cpp` for existing tests for these functions, and add new tests that make sure that msgids with invalid characters are rejected.

IgorSivy:
Hello there, could I try to solve this issue?

goatshriek:
Absolutely! I will go ahead and assign it to you. Please feel free to reach out here or on gitter if you have any issues or questions.

[...]

PR #127, opened 2020-10-20
Msgid format check

IgorSivy:
This pull request resolves issue #114 

Overview of what I've done:
1. Created validate_msgid_format function to prevent msgids with wrong format from being accepted. 
2. Created new error message and code type to indicate wrong encoding errors. 
3. Added 3 tests to check if the new function has correct functionality.


goatshriek:
Excellent work, and thank you for implementing this change!

Issue #128, opened 2020-10-21
implement param to string method


goatshriek:
It is often useful to be able to obtain out a string representation of a given value either for user interface hints, troubleshooting, or operations like value comparison. Such a function is currently implemented for some stumpless data types (like the version struct) but not for others, including the param struct.

Note that this function is _not_ intended for use in the formatting of a log message as this output may be subtly different, for example when escape codes are needed. Rather, it is meant to give an as-is represenation of the param itself (name and value).

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block. You are also free to take a different approach if you feel it is better - this is merely meant as a starting point.

First, read [this specific section](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions) of the development guide devoted to adding new functions. It provides an outline of the work you will need to do beyond the implementation and testing of the function.

Next, add a declaration for your function in the `include/stumpless/param.h` header file. Name the function `stumpless_param_to_string` and document its behavior in the same doxygen format as the others in the same header. Make sure to cover the following points:
 * the general format of the output (suggested: `<name>: <value>`)
 * the supplied argument should be a `const` pointer
 * explicitly state that the result must be freed when it is no longer needed

Now you are ready to add the matching implementation for the function in `src/param.c`. Use the code for existing functions in this file as examples. Be sure to use the length measurements already available in the struct to avoid any `strlen` calls when determining the size of the memory to allocate. Also avoid expensive operations like `sprintf` - this is doable using simple `memcpy` calls and assignments.

Finally, be sure to add tests for the new function. These should go in `test/function/param.cpp` and should achieve full coverage, including for memory allocation failure and NULL argument cases. Again, there are already similar test cases for other functions in this file that you can look at to get a general idea of how the tests should be structured.

hoomaac:
sounds interesting! I'm working on it.

goatshriek:
Great! I've assigned the issue to you to mark it as taken.

PR #134, opened 2020-11-13
add stumpless_param_to_string function

hoomaac:
Implemented param to string function as described in issue #128. It gets the param and return a formatted string like;
 \<name\>:\<value\> with memcpy and assignment.


goatshriek:
This PR looks good to me!

For some reason, Travis CI missed the notification about this PR and failed to build it. I pushed a local branch with the commit to make sure that all CI builds run and diff coverage is good - once these complete I will merge this PR in (barring any issues, of course).

PR #138, opened 2020-11-20
add cmake, vs and ctag files to gitignore

hoomaac:
cmake files and build folder, Visual studio out folder and ctag files are now ignored by git.


.gitignore
```diff
@@ -9,4 +9,31 @@ docs/html/*
 *.so
 *.swp
 .vs/*
+out/*
```
goatshriek: looks like there is a duplicate entry for the out directory - I would recommend removing this one, as the following one has a comment that makes its reason clear.

goatshriek (changes requested):
Just a minor tweak to make to make this a little cleaner, and then it will be good to go!

.gitignore
hoomaac: yes. I didn't see that. I've updated .gitgnore.

goatshriek (approved):

Issue #129, opened 2020-10-21
length checking of the app-name


goatshriek:
The syslog RFC ([RFC 5424](https://tools.ietf.org/html/rfc5424)) specifies a maximum length for many fields of a log message, including the app-name field. Checking the length of supplied app-name arguments will not only help users stay compliant with the standard, but will also allow for efficiencies in the implementation by eventually allowing for the use of fixed-length buffers instead of memory allocated from the heap.

As specified in the RFC, app-name parameters should be checked for their length and rejected if they are more than the allowed length (48 characters). Note that this length does not include the terminating NULL character from the C string representation.

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block. You are also free to take a different approach if you feel it is better - this is merely meant as a starting point.

There are a number of functions that allow an app-name to be supplied - the constructor for entries, the set feature for the app-name of entries, and the default app-name setting for targets, to list a few. Since this check will need to take place in several places, you will be best served by implementing it in a single function and then calling it wherever required.

Declare your new function (something like `validate_app_name_length`) in `include/private/validate.h`, and define it in `src/validate.c`. A similar validation function has already been created for msgids - use it as a reference as needed. After your new function has been defined, add calls to it wherever an app-name is passed as a parameter in the library, which are the following functions:
 * `vstumpless_new_entry` (in `src/entry.c`)
 * `stumpless_set_entry_app_name` (in `src/entry.c`)
 * `stumpless_set_target_default_app_name` (in `src/target.c`)

Update the documentation for each of these functions to specify that the app name length is restricted to be 48 characters or less. The relevant documentation is in a doxygen comment format in the include files corresponding to the source files (`include/stumpless/entry.h` and `include/stumpless/target.h`). Add the note about this to the `@param` descriptions of the app-name parameters.

[...]

PR #139, opened 2020-11-22
add validation on app name length

hoomaac:
implemented validate_app_name_length function. Now app name length is restricted to be 48 characters or less as described in issue #129 .

### DIFF
--- a/include/private/validate.h
+++ b/include/private/validate.h
@@ -61,13 +61,4 @@ bool validate_msgid_format( const char* msgid );
  */
 bool validate_app_name_length( const char* app_name);
 
-
-
-
-
-
-
-
-
-
 #endif /* __STUMPLESS_PRIVATE_VALIDATE_H */
--- a/include/private/validate.h
+++ b/include/private/validate.h
@@ -50,4 +50,24 @@ bool validate_msgid_length(const char* msgid );
  */
 bool validate_msgid_format( const char* msgid );
 
+/**
+ * Checks the length of app name.
+ *
+ * @param the app name
+ *
+ * @return True if the app name is less than allowed length 
+ * (48 characters not including NULL terminating),otherwise
+ * it will return false and raise STUMPLESS_ARGUMENT_TOO_BIG error.
+ */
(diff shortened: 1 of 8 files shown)


src/entry.c
```diff
@@ -686,13 +691,19 @@ vstumpless_new_entry( int facility,
...
+
   app_name_length = &( entry->app_name_length );
   entry->app_name = copy_cstring_with_length( effective_app_name,
                                               app_name_length );
   if( !entry->app_name ) {
     goto fail_app_name;
   }
 
+  
+
```
goatshriek: two more extra lines to remove

include/private/validate.h
```diff
@@ -50,4 +50,24 @@ bool validate_msgid_length(const char* msgid );
...
+
+
+
+
+
+
+
+
+
+
```
goatshriek: The remaining whitespace here is extra; only leave one empty line before the endif.

test/function/entry.cpp
```diff
@@ -548,20 +548,20 @@ namespace {
   }
 
   TEST_F( EntryTest, SetAppName ) {
```
goatshriek: it appears that this function had its indentation inadvertently changed - be sure to revert it (as no changes were made)

test/function/target.cpp
```diff
@@ -23,6 +23,7 @@
 #include <gtest/gtest.h>
 #include <stumpless.h>
 #include "private/target/buffer.h"
+#include "private/validate.h"
```
goatshriek: you can safely remove this include (it is causing your travis builds to fail)

goatshriek (changes requested):
Overall, very good work! There are a few cosmetic whitespace tweaks and an extra include to fix up with this, and then it looks like it will be good to go.

goatshriek (approved):

goatshriek:
These changes look great, thanks for putting this all together! I will merge this in once the Appveyor builds complete. You can ignore the cancelled Travis CI build; they have dropped free support for OSS, and I am still working on a migration to Github Actions to fill the gap.

PR #148, opened 2020-12-10
deprecate ambiguous destructors

goatshriek:
There are two forms of the destructors for entry and element structures:  one that destroys the object itself, and one that destroys the object and all of the ones that it contains. The former is named `destroy_..._only` while the latter is named  `destroy_..._and_contents`. The latter has a synonym named simply `destroy_...`, which does not convey the nuance of its behavior at all. This alias has therefore been deprecated, and will be removed in release 3.0 in order to prevent confusion and misuse of the two forms.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,6 +87,7 @@ jobs:
       - name: Test
         run: |
           make check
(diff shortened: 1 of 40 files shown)


goatshriek:
A drop in coverage is expected in this change, as the deprecated functions are no longer called by any tests.

Issue #162, opened 2021-01-19
check target type in close functions


goatshriek:
There is a specific function for closing each type of target, but it does not check the type of the target before closing. If passed the wrong type of target, this could lead to memory corruption and/or crashing.

While this would be a user error, a check for the target type would reduce the chance of this becoming a problem. A future build option to remove error checks if needed would allow the performance penalty to be removed if users can guarantee they won't violate this.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block. You are also free to take a different approach if you feel it is better - this is merely meant as a starting point.

First, take a look at the `raise_target_incompatible` function in `include/private/error.h`. You will use this function to raise an error when the wrong target type is detected.

You need a string to pass to the error raising function as the error message. Since there isn't an appropriate one existing, you will need to add a new one to the locales. Check out the [documentation on localization](https://github.com/goatshriek/stumpless/blob/latest/docs/localization.md#adding-new-strings) for instructions on how to do this.

Implementing the actual checks is straightforward. For each close function (of the form `stumpless_close_xxx_target`) in the target modules (in the `src/target/` directory), add a check on the target's type and raise an error (and return of course) if it is not the expected type.

Finally, add tests for each modification to make sure that trying to close the wrong target type is caught. For each target type, look at the existing test suite (these are in the `test/function/target/` directory) and add a test for the close test with a different target type. A good starting place is the test that passes `NULL` into the function and checks for an error - copy this test and modify it to test for the new scenario.

ShobhitSamuelSingh:
Hi, I would like to work on this issue.

goatshriek:
Sounds good! I have assigned you so that it's clear to folks that someone is working on this.

ShobhitSamuelSingh:
Thank you for assigning me the issue. Please check if the changes made are correct and I couldn't add test cases because I didn't know how to, please help me with that too.

goatshriek:
Sure thing, I provided comments and more information in #163 for you to take a look at.

goatshriek:
This issue is re-open for anyone that would like to work on it. The pull request referenced above has initial work done by @ShobhitSamuelSingh which is a good starting point if you would prefer not to start from scratch.

PR #163, opened 2021-01-21
Made Changes as per Issue #162

ShobhitSamuelSingh:
Please check if the changes made are correct, and please guide me about how to add test cases.


goatshriek:
It looks like there are some issues with compilation - take a look at your updates to `src/target/stream.c`, which has code outside of a function (starting on line 49) and references a string literal macro as a function (on line 40). Running the `check` target (for example `make check` in a Makefile-based build) on your code will let you verify that everything builds and no existing tests are breaking.

It also appears that you have checked for a buffer target type in every function, instead of the respective type. For example, the check in `src/target/file.c` should make sure that the target type is `STUMPLESS_FILE_TARGET`. You could leave it this way to validate that your tests fail, but just be sure you do fix it afterwards to get the tests passing.

Regarding testing - I'll walk you through the buffer target test first. Once you get this working, you would do essentially the same thing for the other target types.

First, look for the test of the `stumpless_close_buffer_target` function that passes in a NULL target. For the buffer targets, this is at line 271 of `test/function/target/buffer.cpp`. Make a copy this test directly below the existing one, and name the new one something fitting, perhaps `TEST(BufferTargetCloseTest, WrongTargetType )`. Next, create a new target that is _not_ a buffer target; I recommend using `stumpless_open_stdout_target` for this. Pass this as the argument to the close function (instead of NULL). Change the expected error id to be `STUMPLESS_TARGET_INCOMPATIBLE`. Finally, be sure you close the target you created (you can just use `stumpless_close_target` for this). You can then run the `check` target to make sure that all of the tests pass, including your new one.

goatshriek:
This pull has been marked as stale since there have been no updates in the last two months. It will be closed after another month of inactivity.

ShobhitSamuelSingh:
I am so sorry for the late reply, but this is just because my exams were going on and they have finally ended now.
I would like to request you to please close this pull request as I am still a beginner and it's getting slightly difficult for me to understand. Thank you so much for all your guidance. I hope I can contribute to this project in some other way afterward.

goatshriek:
No problem at all! I hope that you picked up some useful tidbits while working on this, and that you come back once you've got a better handle on things!

Issue #156, opened 2020-12-28
check italian (it-IT) localization


goatshriek:
Localization for Italian was added in #155 by the library's main author (@goatshriek). However, the author's command of the Italian language is not very strong, and as such the translations should be checked for accuracy and general correctness.

This is a straightforward review - simply look at `include/private/config/locale/it-it.h` and submit a pull request with updates and corrections.

PR #168, opened 2021-03-24
Fix translations #156

beninidavide:
Fix some italian translations


beninidavide:
Hi @goatshriek, can I do anything to fix the failing check?

goatshriek:
Hi @Boccolo! There's no need to worry about the failure here, it is a result of the nuances of running CI builds against external PRs.

Thanks very much for correcting this localization! Clearly I have a lot of studying left to do! 😜

PR #172, opened 2021-04-13
Added app name check, docs, and tests

### LINKED ISSUE #130: format checking of app-name
The syslog RFC ([RFC 5424](https://tools.ietf.org/html/rfc5424)) specifies that a number of fields only include printable ASCII characters (decimal values 33 to 126 inclusive), including the app-name. Checking the characters of supplied app-name arguments will not only help users stay compliant with the standard, but will also provide a point of mitigation for encoding bugs and vulnerabilities.

## General Approach

There are a number of functions that allow an app-name to be supplied - the constructor for entries, the set feature for the app-name of entries, and the default app-name setting for targets. Since this check will need to take place in several places, you will be best served by calling a single function wherever required.

But good news: a function that does this already exists! It is in `src/validate.c` and is named `validate_msgid_format`. You just need to do some refactoring of this function to make it more generalized to support app-names (and other fields in the future). This amounts to renaming the function to something more general (perhaps `validate_printable_ascii`?) and changing invocations of it to this new name. An easy approach to take here is to rename the definition and declaration, and then follow the resulting compile errors to find the invocations. Or do a `grep`! Or some other find/replace method, it is up to you.

After this refactoring, add calls to the function wherever an app-name is passed as a parameter in the library, which are the following functions:
 * `vstumpless_new_entry` (in `src/entry.c`)
 * `stumpless_set_entry_app_name` (in `src/entry.c`)
 * `stumpless_set_target_default_app_name` (in `src/target.c`)

Update the documentation for each of these functions to specify that the app name is restricted to specific characters. The relevant documentation is in a doxygen comment format in the include files corresponding to the source files (`include/stumpless/entry.h` and `include/stumpless/target.h`). Add the note about this to the `@param` descriptions of the app-name parameters.

Don't forget to add tests for your new functionality! These should go alongside the tests for the functions that have app-names - check `test/function/target.cpp` and `test/function/entry.cpp` for existing tests for these functions, and add new tests that make sure that app-names with invalid characters are rejected.

CuckooEXE:
Fixed #130, modified the function name, updated references and documentation, added new tests. I wasn't too confident on the tests, it's 100% pass, but I'm not sure if I really am testing it correctly. 

Let me know if I can do anything to improve the PR!

### DIFF
--- a/src/validate.c
+++ b/src/validate.c
@@ -38,11 +38,11 @@ bool validate_msgid_length(const char* msgid ) {
   return validation_status;
 }
 
-bool validate_printable_ascii( const char* app_name ) {
-  size_t app_name_length = strlen( app_name );
+bool validate_printable_ascii( const char* str ) {
+  size_t str_length = strlen( str );
 
-  for (size_t i = 0; i < app_name_length; i++) {
-    if (app_name[i] < 33 || app_name[i] > 126) {
+  for (size_t i = 0; i < str_length; i++) {
+    if (str[i] < 33 || str[i] > 126) {
       raise_invalid_encoding(L10N_FORMAT_ERROR_MESSAGE("app name"));
       return false;
     }
--- a/src/validate.c
+++ b/src/validate.c
@@ -38,18 +38,19 @@ bool validate_msgid_length(const char* msgid ) {
   return validation_status;
 }
 
-bool validate_msgid_format( const char* msgid ) {
-  size_t msgid_length = strlen( msgid );
+bool validate_printable_ascii( const char* app_name ) {
+  size_t app_name_length = strlen( app_name );
 
-  for (size_t i = 0; i < msgid_length; ++i) {
-    if (msgid[i] < 33 || msgid[i] > 126) {
-      raise_invalid_encoding(L10N_FORMAT_ERROR_MESSAGE("msgid"));
+  for (size_t i = 0; i < app_name_length; i++) {
+    if (app_name[i] < 33 || app_name[i] > 126) {
+      raise_invalid_encoding(L10N_FORMAT_ERROR_MESSAGE("app name"));
       return false;
     }
   }
 
   return true;
 }
+
 
 bool validate_app_name_length( const char* app_name ) {
     size_t max_app_name_length = 48;
--- a/test/function/entry.cpp
+++ b/test/function/entry.cpp
@@ -877,13 +877,22 @@ namespace {
     EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
   }
 
-  TEST_F( EntryTest, SetAppNameRejected ) {
+  TEST_F( EntryTest, SetAppNameRejectedLen ) {
      const struct stumpless_entry* result;
      const struct stumpless_error* error;
 
     result = stumpless_set_entry_app_name( basic_entry, "it-is-gonna-fail-because-length-exceeded-max-allowed" );
 
     EXPECT_ERROR_ID_EQ( STUMPLESS_ARGUMENT_TOO_BIG );
+  }
+
+  TEST_F( EntryTest, SetAppNameRejectedFormat ) {
+     const struct stumpless_entry* result;
+     const struct stumpless_error* error;
+
+    result = stumpless_set_entry_app_name( basic_entry, "test-appname-wro\ng-format" );
+
+    EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
   }
 
   TEST_F( EntryTest, AppNameRejected ) {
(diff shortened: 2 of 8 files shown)


src/validate.c
```diff
@@ -38,19 +38,20 @@ bool validate_msgid_length(const char* msgid ) {
   return validation_status;
 }
 
-bool validate_msgid_format( const char* msgid ) {
-  size_t msgid_length = strlen( msgid );
+bool validate_printable_ascii( const char* app_name ) {
```
goatshriek: Change the parameter name here to match what you renamed it to in the `validate.h` header so that it is more general (`str`).

src/validate.c
```diff
@@ -38,19 +38,20 @@ bool validate_msgid_length(const char* msgid ) {
...
+  size_t app_name_length = strlen( app_name );
```
goatshriek: Change this variable name to match the parameter name (`str_length`).

test/function/entry.cpp
```diff
@@ -1188,6 +1188,21 @@ namespace {
...
+    struct stumpless_entry *bad_stump;
+    const struct stumpless_error *error;
+
+
+    const char bad_msgid[] = { 'b', 'a', 'd', '_', 0x5, 'n', 'a', 'm', 'e', '\0' };
+    bad_stump = stumpless_new_entry( STUMPLESS_FACILITY_USER,
+                                     STUMPLESS_SEVERITY_INFO,
+                                     basic_app_name,
+                                     bad_msgid,
+                                     basic_message );
```
goatshriek: Here you should pass the bad value as the `app_name` and not the `msgid` parameter.

test/function/entry.cpp
```diff
@@ -1188,6 +1188,21 @@ namespace {
...
+
+    EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
```
goatshriek, line 1210: Consider adding another check to make sure that the return value is `NULL` here using `EXPECT_NULL`. It hasn't been done in other tests unfortunately, but is certainly worthwhile to add for this one (and the others) as you make your adjustments.

test/function/target.cpp
```diff
@@ -162,6 +162,22 @@ namespace {
...
+    bad_stump = stumpless_new_entry( STUMPLESS_FACILITY_USER,
+                                     STUMPLESS_SEVERITY_INFO,
+                                     "basic-app-name",
+                                     bad_msgid,
+                                     "basic_message" );
+
+    EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
+  }
+
+
```
goatshriek: This test should be removed and re-written as a non-fixture test to cover `stumpless_set_target_default_app_name` - it is currently identical to the previous test. Refer to the similar test for the default msgid which you can find in the same file at [line 639](https://github.com/goatshriek/stumpless/blob/b0e8b2ff6e7a669060147538ef16e03f76f48545/test/function/target.cpp#L639).

goatshriek (changes requested):
Things look like they're shaping up well! There are a few more additions to finish this up, but I think you will be able to make them without much trouble.

It looks like you've finished the refactoring of the `msgid` checks, but still need to add the checking to the `app_name` uses in the three functions listed in the issue. Once you adjust your tests to cover the `app_name` in those three functions, they will verify that you're covering them. I recommend making the test updates first so that you see them fail, and then adding the validations so that you see them resolve.

In addition to the test adjustments noted in the per-line comments, also make sure that you add a test for the `stumpless_set_entry_app_name` function, which is currently uncovered. You can find a reference test on [line 871 of `test/function/entry.cpp`](https://github.com/goatshriek/stumpless/blob/b0e8b2ff6e7a669060147538ef16e03f76f48545/test/function/entry.cpp#L871) for the equivalent msgid functionality - you should be able to copy it as a starting point.

CuckooEXE:
Made the changes you suggested! Now `test/target.cpp` have (semi-)identitical tests for `stumpless_set_target_default_msgid` and `stumpless_set_target_default_app_name` to change poorly formatted names, and `test/entry.cpp` has tests for `stumpless_set_entry_msgid` and `stumpless_set_entry_app_name`.

test/function/target.cpp
```diff
@@ -162,6 +162,20 @@ namespace {
...
+    bad_stump = stumpless_new_entry( STUMPLESS_FACILITY_USER,
+                                     STUMPLESS_SEVERITY_INFO,
+                                     "basic-app-name",
+                                     "bad\nmsgid",
+                                     "basic_message" );
+
+    EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
+  }
+
+
```
goatshriek: This should be removed - it is duplicative of another msgid test that already exists.

goatshriek:
The tests look good, and as expected they are failing (which is good)! Now you just need to add calls to the `validate_printable_ascii` function for the app_name parameters in the three functions, and they should all resolve. You don't need to worry about resolving the code analysis check - it won't run for external pull requests.

CuckooEXE:
Sorry! You beat me to the punch, I realized I didn't push up the most important changes. Should all be there now : ) 100% tests passed

goatshriek:
No worries, you were already on top of it! Everything looks good to go - just remove the duplicative test in the target suite (lines 165-178, see above comment) and I'll merge this in.

test/function/target.cpp
CuckooEXE: Done!

Issue #146, opened 2020-12-08
implement element to string function


goatshriek:
It is often useful to be able to obtain a string representation of a given value either for user interface hints, troubleshooting, or operations like value comparison. Such a function is currently implemented for some stumpless data types (like the version struct) but not for others, including the element struct.

Note that this function is _not_ intended for use in the formatting of a log message as this output may be subtly different, for example when escape codes are needed. Rather, it is meant to give an as-is represenation of the element itself (name and all params).

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block. You are also free to take a different approach if you feel it is better - this is merely meant as a starting point.

Before you start, take a peek at bb30c61b82be4fa031a719c8457b562981fef8cf which implements a similar functionality for params. You will need to use the function implemented here, and the basics of this change will also apply to this one as well.

First, read [this specific section](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions) of the development guide devoted to adding new functions. It provides an outline of the work you will need to do beyond the implementation and testing of the function.

Next, add a declaration for your function in the `include/stumpless/element.h` header file. Name the function `stumpless_element_to_string` and document its behavior in the same doxygen format as the others in the same header. Make sure to cover the following points:
 * the general format of the output (suggested: `<name>` (with no params), `<name>:[param_1_to_string,param_2_to_string,etc.]` with params)
 * the supplied argument should be a `const` pointer
 * explicitly state that the result must be freed when it is no longer needed

Now you are ready to add the matching implementation for the function in `src/element.c`. Use the code for existing functions in this file as examples. You will also need to use `stumpless_param_to_string` to get the string representation of any params in the element. Be sure that there is no memory leak in your result, and try to avoid expensive operations like `sprintf` - you should be able to do this with `memcpy` calls and assignments, and perhaps `strlen` calls. You may consider refactoring some common portions of `stumpless_param_to_string` into separate private functions that can be used in both places for efficiency, but don't worry if you don't want to go this far - it can be turned into a follow-on first issue for others to tackle.

[...]

Issue #177, opened 2021-05-01
format checking of param names


goatshriek:
The syslog RFC ([RFC 5424](https://tools.ietf.org/html/rfc5424)) specifies that a number of fields only include printable ASCII characters (decimal values 33 to 126 inclusive), including structured data identifiers such as param names. In addition to being limited to printable ASCII, identifiers have even more restricted characters that are not allowed. Checking the characters of supplied identifiers will not only help users stay compliant with the standard, but will also provide a point of mitigation for encoding bugs and vulnerabilities.

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

There are a number of functions that allow a param name to be supplied: the constructor for params, accessor and mutator functions for entries and elements. Since this check will need to take place in several places, you will be best served by calling a single function wherever required.

A validation function already exists to check that a string only contains printable ASCII characters: `validate_printable_ascii` in `src/validate.c`. You will need to define a new function in the same file for structured data identifiers, adding the restriction that the string also not contain these characters and raising an encoding error if it does:
 * an equal sign `=`, decimal value 61
 * a closing brace `]`, decimal value 93
 * a double-quote character `"`, decimal value 34

After this addition, add calls to your new function in functions where a param name is passed as a parameter in the library, which are listed below. For functions where the parameter is used for a search, be sure to check the name _before_ acquiring any locks to avoid deadlock and improve performance.
 * `stumpless_new_param` (in `src/param.c`) This function is used by all others that create a new param, and so explicit checks are not needed in those.
 * `stumpless_set_param_name` (in `src/param.c`) This function is used by all others that modify a param name, and so explicit checks are not needed in those.
 * `stumpless_element_has_param` (in `src/element.c`)
 * `stumpless_get_param_by_name` (in `src/element.c`)
 * `stumpless_get_param_index` (in `src/element.c`)
 * `stumpless_get_param_name_count` (in `src/element.c`)
 * `stumpless_get_param_value_by_name` (in `src/element.c`)
 * `stumpless_get_entry_param_by_name` (in `src/entry.c`)
[...]

PR #180, opened 2021-05-28
Implement stumpless_element_to_string function

northernSage:
This addresses #146 

- [x] add declaration to `element.h`
- [x] implement `stumpless_element_to_string` in element.c
- [x] add tests

### DIFF
--- a/include/stumpless/element.h
+++ b/include/stumpless/element.h
@@ -762,6 +762,35 @@ stumpless_set_param_value_by_name( struct stumpless_element *element,
                                    const char *name,
                                    const char *value );
 
+/**
+ * Returns name and params from element as a formatted string.
+ * The character buffer should be freed when no longer is needed by the caller.
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe. A mutex is used to coordinate the read of the
+ * element with other accesses and modifications.
+ *
+ * **Async Signal Safety: AS-Unsafe lock heap**
+ * This function is not safe to call from signal handlers due to the use of a
+ * non-reentrant lock to coordinate access and the use of memory management
+ * functions to create the result.
+ *
+ * **Async Cancel Safety: AC-Unsafe lock heap**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, due to the use of a lock that could be left locked as well as
+ * memory management functions.
+ *
+ * @since release v2.1.0
+ *
+ * @param element The element to get the name and params from.  
+ *
+ * @return The formatted string of <name> or <name>:[param1,...] if no error is encountered.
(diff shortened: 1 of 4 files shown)


northernSage:
I have created the draft to open what I have done so far for review (early feedback makes for a less painful process :bowtie: ) In case anyone has suggestions on how to improve the PR feel free to let me know.

goatshriek:
Looks like you're on a good path! Right now the only comment I have is to be sure that you free the result of param_to_string after you memcpy it into the new buffer, otherwise it will become a memory leak.

northernSage:
Thanks for the heads up, will get it sorted and finished the PR in the next few days 👍🏻

northernSage:
This should be ready for review, feel free to suggest any other improvements or corrections @goatshriek

goatshriek:
Looks like you're pretty much there, awesome work on putting this together! My apologies for the delay in running the CI actions, I forgot that they must be manually approved now. Just a few things left to clean up:

The most difficult issue is that the Microsoft compiler doesn't appear to support the non-compile-time constant size in the array declaration on line 525 of `src/element.c`, which you probably already noticed from the failed actions. You could allocate that array from the heap instead, append the strings as you iterate through the params one by one instead of storing all of the strings first, or take a different approach entirely, it's up to you.

You'll need to add an entry to the `src/windows/stumpless.def` file so that the function is exported in the DLL, otherwise the Windows tests will continue to fail even after the compilation issue is resolved. Just tack the new function name on to the end of the file with the next available ordinal number and you should be all set.

The branch covering elements without any params is currently uncovered - adding another test almost identical to the one starting on line 712 of `test/function/element.cpp` but with the `basic_element` fixture element instead of the `element_with_params` one should cover this.

You can ignore the sonarcloud failure - it won't run for "external" pull requests at the moment.

northernSage:
Thanks for pointing out the issues in such detail, it helped a lot! I believe this is ready for another go :bowtie:

goatshriek:
Looks great! the mac failures look like a false alarm - I will give this one more check in a few hours and merge it in!

northernSage:
From #146: 

> You may consider refactoring some common portions of stumpless_param_to_string into separate private functions that can be used in both places for efficiency...

Should I open a follow up issue for this? I would be willing to work on it after #177 @goatshriek

goatshriek:
If you're interested, absolutely! There is a detailed walkthrough of making/measuring performance optimizations in [`docs/benchmark.md`](https://github.com/goatshriek/stumpless/blob/latest/docs/benchmark.md) that walks through the general flow used to make/test speed improvements.

There's no need to feel obligated though; I can always write another `good first issue` for the improvements as well. Totally up to you!

PR #184, opened 2021-06-13
Implement format checking of param names

northernSage:
following #177

- [x] implement `validate_param_name`
- [x] add `validate_param_name` declaration to `validate`.h
- [x] update docs for `stumpless_new_param` and `stumpless_set_param_name`

validate param names in:
- [x] `stumpless_new_param` 
- [x] `stumpless_set_param_name`
- [x] `stumpless_element_has_param` 
- [x] `stumpless_get_param_by_name`
- [x] `stumpless_get_param_index`
- [x] `stumpless_get_param_name_count` 
- [x] `stumpless_get_param_value_by_name` 
- [x] `stumpless_get_entry_param_by_name`
- [x] `stumpless_get_entry_param_value_by_name`

Write tests with invalid param names for :
- [x] `stumpless_new_param` 
- [x] `stumpless_set_param_name`
- [x] `stumpless_element_has_param` 
- [x] `stumpless_get_param_by_name`
- [x] `stumpless_get_param_index`
- [x] `stumpless_get_param_name_count` 
- [x] `stumpless_get_param_value_by_name` 
- [x] `stumpless_get_entry_param_by_name`
- [x] `stumpless_get_entry_param_value_by_name`

### DIFF
--- a/include/private/validate.h
+++ b/include/private/validate.h
@@ -61,4 +61,15 @@ bool validate_printable_ascii( const char* str );
  */
 bool validate_app_name_length( const char* app_name);
 
+/**
+ * Checks that the passed in param name contains only ASCII characters and
+ * also does not contain any of the following characters: '=',']','"'.
+ *
+ * @param the param name.
+ *
+ * @return True if the param name has the correct format, otherwise
+ * it will return false and raise STUMPLESS_INVALID_ENCODING error.
+ */
+bool validate_param_name( const char* str);
+
 #endif /* __STUMPLESS_PRIVATE_VALIDATE_H */
--- a/include/stumpless/param.h
+++ b/include/stumpless/param.h
@@ -207,7 +207,7 @@ stumpless_get_param_value( const struct stumpless_param *param );
  * This function is not safe to call from threads that may be asynchronously
  * cancelled, due to the use of memory management functions.
  *
- * @param name The name of the new param.
+ * @param name The name of the new param. Restricted to printable ASCII characters different from '=', ']' and '"'.
  *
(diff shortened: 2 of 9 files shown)


northernSage:
I have been slowly chipping away at this when I find the time for a few days now, feel free to let me now if something doesn't look good @goatshriek 👍🏻  We are getting a lot of repeated code in the tests, so I wonder if there's a better approach here. I might refactor this later on to clean up a bit, perhaps extract the three invalid test strings ("par]am", "par\"am" and "par=am") into an array and loop over it asserting null/false is returned and the proper exceptions are raised (?) That would look a bit more clean, I suppose.

goatshriek:
My apologies for the delayed response, somehow I missed that this was ready for review. It is excellent work, thank you for putting so much time into it!

I see what you mean about the repetition in the test code. Normally what I would do is create a helper macro somewhere in `include/test/helper` that I could then use. In looking through the development documentation though, I see that there really isn't anything on the test suites or setup that would suggest that as a potential solution, which is probably something I should fix. I'll make a note to add that soon. Do you have any suggestions on what info should be go into that documentation refresh?

northernSage:
> My apologies for the delayed response, somehow I missed that this was ready for review.

Don't worry about it @goatshriek 👍🏻 

>  Do you have any suggestions on what info should be go into that documentation refresh?

So far all questions I had while working on my PRs were covered in detail by the docs and example code, so I would say the documentation is already excellent as is. Adding the section about the helper macro would indeed be nice, though. Probably make a follow up issue addressing the clean-up of test code after that (?) I would be willing to work on it.

Issue #120, opened 2020-09-20
implement spanish (es-ES) localization


goatshriek:
Localization has been added to the library, and Spanish (es-ES) has since been added with stub values. However, these messages are not useful and need to be replaced with actual Spanish to be of any use.

This will be straightforward for someone that speaks the language, even if you have very little knowledge of C. You will simply need to update the strings in `include/private/config/locale/es-es.h` with Spanish strings (in UTF-8 encoding). If you want to see what the English message is, check the corresponding values in `include/private/config/locale/en-us.h`.

ulisesten:
Does this need help yet?

goatshriek:
It is all yours if you would like to work it!

ulisesten:
ok, let's do it

PR #185, opened 2021-07-04
Implement spanish (es-ES) localization (#120)

ulisesten:
Implementation for spanish (es-ES) has been added as it  was required on issue #120.
ChangeLog.md was modified to add the changes.


ChangeLog.md
```diff
@@ -12,6 +12,7 @@ fixes, check out the
 ## [2.1.0] - 2021-05-16
 ### Added
  - Custom function logging targets.
+ - Localization for Spanish (es-ES :mx)
```
goatshriek: I believe the flag character needs at least a trailing colon, or perhaps should be `:mexico:` [instead](https://emojipedia.org/flag-mexico/)? As is, it does not render as a flag.

Also, in addition to updating the Changelog, consider updating the Key Features section of the [README](https://github.com/goatshriek/stumpless#key-features) with the new flag as well!

include/private/config/locale/es-es.h
```diff
@@ -20,169 +20,171 @@
...
 #  define L10N_STRING_LENGTH_ERROR_CODE_TYPE \
-"STRING LENGTH ERROR CODE TYPE"
+"el largo de la cadena infractora"
 
 #  define L10N_FORMAT_ERROR_MESSAGE(ARG) \
-"invalid " ARG " format"
+"formato " ARG " no válido"
 
 #  define L10N_INVALID_TARGET_TYPE_ERROR_MESSAGE \
 "INVALID TARGET TYPE ERROR MESSAGE"
```
goatshriek: Looks like this is the last one that needs to be updated.

goatshriek (changes requested):
Thanks very much for putting this together, it looks great! Just a few minor changes to make, and it will be good to go!

ChangeLog.md
ulisesten: Thanks. I'll change the tag to :mexico: and add it to README.md

ulisesten: but (es-ES) means Spain spanish, a think that the tag would be :es:,  because :mexico: would be (es-MX). The translation is the same, so there will be no problem with it

goatshriek: Ah, great point!

Issue #187, opened 2021-09-01
implement format checking of element names


goatshriek:
The syslog RFC ([RFC 5424](https://tools.ietf.org/html/rfc5424)) specifies that a number of fields only include printable ASCII characters (decimal values 33 to 126 inclusive), including structured data identifiers such as element names. In addition to being limited to printable ASCII, identifiers have other restricted characters that are also not allowed. Checking the characters of identifiers will not only help users stay compliant with the standard, but will also provide a point of mitigation for encoding bugs and vulnerabilities.

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

A similar issue for param names has already been tackled, and is a great place to review if you want some more preparation material. Check out the excellent work by Gabriel Fioravante in #177 and #184 for inspiration.

There are a number of functions that allow a element name to be supplied: the constructor for elements, accessor and mutator functions for entries and elements. Since this check will need to take place in several places, you will be best served by calling a single function wherever required.

A validation function already exists to check that a string contains only allowed characters: `validate_param_name` in `src/validate.c`. You will need to either copy this function and change the copy for element names or refactor the current one (via rename) to support both.

After this change, add calls to your new function wherever an element name is passed as a parameter in the library, which are listed below. For functions where the element name is used for a search, be sure to check the name _before_ acquiring any locks to avoid deadlock and improve performance.
 * `stumpless_new_element` (in `src/element.c`) This function is used by all others that create a new element, and so explicit checks are not needed in those.
 * `stumpless_set_element_name` (in `src/element.c`) This function is used by all others that modify an element name, and so explicit checks are not needed in those.
 * `stumpless_entry_has_element` (in `src/entry.c`)
 * `stumpless_get_element_by_name` (in `src/entry.c`)
 * `stumpless_get_element_index` (in `src/entry.c`)
 * `stumpless_get_entry_param_by_name` (in `src/entry.c`)
 * `stumpless_get_entry_param_value_by_name` (in `src/entry.c`)

[...]

PR #188, opened 2021-10-07
Introduced checking for element name

anindyasen:
1. Element name check utility is introduced.
2.  Check is introduced in all the function mentioned in the bug #187 
3. Test cases for all the function is written

### DIFF
--- a/src/validate.c
+++ b/src/validate.c
@@ -77,3 +77,14 @@ bool validate_param_name( const char* str) {
   return true;
 }
 
+bool validate_element_name( const char* str) {
+  size_t str_length = strlen( str );
+  for (size_t i = 0; i < str_length; i++) {
+    if (str[i] < 33 || str[i] > 126 || str[i] == '=' || str[i] == ']' || str[i] == '"') {
+      raise_invalid_encoding(L10N_FORMAT_ERROR_MESSAGE("invalid element name"));
+      return false;
+    }
+  }
+
+  return true;
+}
--- a/src/validate.c
+++ b/src/validate.c
@@ -69,7 +69,7 @@ bool validate_param_name( const char* str) {
   size_t str_length = strlen( str );
   for (size_t i = 0; i < str_length; i++) {
     if (str[i] < 33 || str[i] > 126 || str[i] == '=' || str[i] == ']' || str[i] == '"') {
-      raise_invalid_encoding(L10N_FORMAT_ERROR_MESSAGE("invalid param name"));
+      raise_invalid_encoding(L10N_FORMAT_ERROR_MESSAGE("param"));
       return false;
     }
   }
@@ -81,7 +81,7 @@ bool validate_element_name( const char* str) {
   size_t str_length = strlen( str );
   for (size_t i = 0; i < str_length; i++) {
     if (str[i] < 33 || str[i] > 126 || str[i] == '=' || str[i] == ']' || str[i] == '"') {
-      raise_invalid_encoding(L10N_FORMAT_ERROR_MESSAGE("invalid element name"));
+      raise_invalid_encoding(L10N_FORMAT_ERROR_MESSAGE("element"));
       return false;
     }
   }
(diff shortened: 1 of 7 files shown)


anindyasen:
Hi @goatshriek, Can you please check and let me know why this test are failing?
I ran make check before commit and it was working perfectly.

Thank you.

src/validate.c
```diff
@@ -77,3 +77,14 @@ bool validate_param_name( const char* str) {
   return true;
 }
 
+bool validate_element_name( const char* str) {
+  size_t str_length = strlen( str );
+  for (size_t i = 0; i < str_length; i++) {
+    if (str[i] < 33 || str[i] > 126 || str[i] == '=' || str[i] == ']' || str[i] == '"') {
+      raise_invalid_encoding(L10N_FORMAT_ERROR_MESSAGE("invalid element name"));
```
goatshriek: For this call, just use the name of the portion that is invalid (`"element"`) as the argument instead of the full `"invalid element name"` string - the localization libraries will add the appropriate language around it saying that it is invalid. This is an issue I missed during review of the last change, which is why you see it in existing code.

test/function/element.cpp
```diff
@@ -1007,6 +1026,30 @@ namespace {
...
+
+    result = stumpless_set_element_name( element, "eleme]nt");
+    EXPECT_NULL( result );
+    EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
+
+    result = stumpless_set_element_name( element, "element\"");
+    EXPECT_NULL( result );
+    EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
+    
+    stumpless_destroy_element( element );
```
goatshriek: This line is causing the test failures that you are seeing, because it uses a deprecated function. `stumpless_destroy_element` has been deprecated because it doesn't make it clear whether the params within the element are being destroyed or not, which is a recipe for memory errors. Changing this to the explicit (and long) `stumpless_destroy_element_and_contents` function should resolve the issue.

The CI builds check the test log for deprecation warnings, which is why you see all of your tests pass and yet the build still fails: there is a deprecation message that must be resolved.

goatshriek (changes requested):
Overall this looks great, thanks for putting it together! I've noted two tweaks for you to make, one of which was causing the failures of the CI jobs. These haven't made it to the coverage report yet, but in reviewing your tests I believe you will have everything covered. Once these are resolved and the coverage is confirmed I will merge this in and apply the hacktoberfest acceptance label (just to make sure).

The code analysis CI job will continue to fail. You can ignore this as it's an access issue as opposed to anything you've done in your changes.

src/validate.c
anindyasen: ok will make it element. Also for param will make it "param".

test/function/entry.cpp
```diff
@@ -566,6 +566,23 @@ namespace {
     EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
   }
 
+  TEST_F( EntryTest, GetParamByNameInvalidElementName ) {
+    const struct stumpless_param *result;
+    const struct stumpless_error *error;
+
+    result = stumpless_get_entry_param_by_name( basic_entry, "ele=ment", "par=am" );
```
goatshriek: for each of these test calls, you'll need to make the param name valid, and just the element name invalid. As is, the check for the invalid element name is not happening, as the param name is failing first. This is why some lines of your pull are not being covered by the tests.

test/function/entry.cpp
```diff
@@ -648,6 +665,23 @@ namespace {
     EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
   }
 
+  TEST_F( EntryTest, GetParamValueByNameInvalidElementName ) {
+    const char *result;
+    const struct stumpless_error *error;
+
+    result = stumpless_get_entry_param_value_by_name( basic_entry, "e=name", "par=am" );
```
goatshriek: The param name needs to be made valid here as well in order to cover the intended execution path.

goatshriek (changes requested):
Looks like the coverage report has exposed one more issue with the tests that I missed - making the tweaks to the test calls for the param to be valid should give you full coverage.

PR #192, opened 2021-10-10
Check for valid element name length

anindyasen:
1. Introduced ```validate_element_name_length``` function.
2. Using the above fundtiont to check element name length.
3. Writing test function to test the new code.


goatshriek:
Excellent work, thanks for putting this together, especially so quickly!

PR #194, opened 2021-10-25
Add implementation for Czech language

palike608:
I hope my translation will be fine. :)


include/private/config/locale/cz-cz.h
```diff
@@ -0,0 +1,196 @@
...
+"element s daným jménem se už vyskytuje v tomto vstupu"
+
+#  define L10N_ELEMENT_NOT_FOUND_ERROR_MESSAGE \
+"není možné najít element se specifikovanou charakteristikou"
+
+#  define L10N_ERRNO_ERROR_CODE_TYPE \
+"volání vrátilo číslo chyby (errno)"
+
+#  define L10N_FILE_OPEN_FAILURE_ERROR_MESSAGE \
+"FILE OPEN CHYBOVÉ HLÁŠENÍ"
```
goatshriek: For those that are in all capital letters, the ideal translation would be one of the phrase in the [English locale](https://github.com/goatshriek/stumpless/blob/latest/include/private/config/locale/en-us.h), instead of the string symbol itself.

These placeholders do exist in other translations for strings that were added after the initial translation, but since this is a new translation it would be ideal to use the more descriptive phrase instead of the symbol name.

goatshriek (changes requested):
These changes look good, thank you for putting them together! I have one minor comment on the translations below, which should be a simple change.

Also, please merge in the other pull requests you have opened to this single branch (`patch-1`) so that they can be bundled together into a single PR for testing. I'll close the others referencing this PR as the main one to merge to.

include/private/config/locale/cz-cz.h
palike608: Ok, i will add anothers edited files into this pull request and i will reedit phrase with all capital letters.

goatshriek (approved):

goatshriek:
Excellent work! Pending the passing of the tests (except for the code analysis one which will fail no matter what), I will merge these in when I have some time in the next 24 hours.

palike608:
thank you for cooperation :)

PR #201, opened 2021-12-15
Documented NOWAIT

Jk-bit:
Documented the use of NOWAIT option in stumpless. Since no child processes are spun the NOWAIT option is essentially on and no implementation is needed of the same. Documentation is at line 61 of `include/stumpless/option.h`


goatshriek:
Looks good, thanks for putting this together!

PR #202, opened 2021-12-17
Implement perror

Jk-bit:
Implemented PERROR option for stumpless. Defined function `stumpless_entry_to_errorstream` in file `src/target.c` on `line 782`
and defined the option `STUMPLESS_OPTION_PERROR` in file `include/stumpless/option.h`. Changes done in `stumpless_add_entry` function in file `src/target.c`. Added test case in `test/function/target.cpp.

Steps to test : 
`cmake --rebuild .`
`cmake --build . --target check`
`./function-test-target`

Check WithPerror category, a log is printed on the terminal with timestamp and message test message with perror
Attached is the image of the output
image


test/function/target.cpp
```diff
@@ -758,6 +780,63 @@ namespace {
     stumpless_close_buffer_target( target );
   }
 
+  TEST( WithPerror, Perror) {
+    struct stumpless_target *target;
+    struct stumpless_target *target_result;
+    char buffer[300];
+    char message_buffer[300];
+    int result;
+    //std::cmatch matches;
```
goatshriek: remove all commented out code

test/function/target.cpp
```diff
@@ -758,6 +780,63 @@ namespace {
...
+    target = stumpless_open_buffer_target( "test target",
+                                           buffer,
+                                           sizeof( buffer ) );
+    ASSERT_TRUE( target != NULL );
+
+    result = stump( "test message without perror" );
+    EXPECT_NO_ERROR;
+    EXPECT_GE( result, 0 );
+
+    stumpless_read_buffer( target, message_buffer, 300 );
```
goatshriek, line 799: follow all calls to `stumpless_read_buffer` with calls to `TestRFC5424Compliance` to verify that the contents comply with formatting standards

src/target.c
```diff
@@ -108,61 +111,107 @@ stumpless_add_entry( struct stumpless_target *target,
   }
 
   clear_error(  );
+ 
+  if( stumpless_get_option( target, STUMPLESS_OPTION_PERROR ) == STUMPLESS_OPTION_PERROR ){
+	// setting is_log_formatted true concludes that STUMPLESS_OPTION_PERROR option is
+	// set, thus only checking this flag is enough for further checks
+	// NOTE : Calling the format_entry twice because if the buffer is for target who
+	// require unformatted entry would need not to call format_entry.
+	is_log_formatted = true;
```
goatshriek: this boolean is unnecessary - you can simply check for whether the buffer is NULL further down in the function, rather than this separate flag

src/target.c
```diff
@@ -90,7 +92,8 @@ stumpless_add_entry( struct stumpless_target *target,
   struct strbuilder *builder;
   size_t builder_length;
   const char *buffer;
```
goatshriek: you will need to initialize the buffer pointer to NULL to serve as an indicator (instead of using a separate boolean)

src/target.c
```diff
@@ -108,61 +111,107 @@ stumpless_add_entry( struct stumpless_target *target,
...
+	// setting is_log_formatted true concludes that STUMPLESS_OPTION_PERROR option is
+	// set, thus only checking this flag is enough for further checks
+	// NOTE : Calling the format_entry twice because if the buffer is for target who
+	// require unformatted entry would need not to call format_entry.
+	is_log_formatted = true;
+  	builder = format_entry( entry, target );
+  	if( !builder ) {
+    		return -1;
+  	}
+  	buffer = strbuilder_get_buffer( builder, &builder_length );
```
goatshriek, line 124: You can simply do the write to the error stream directly from here, instead of in the logging call for each type of target. This will significantly reduce the amount of duplicated code, and make it easier for you to get full test coverage.

src/target.c
```diff
@@ -729,3 +778,11 @@ unsupported_target_is_open( const struct stumpless_target *target ) {
   raise_target_unsupported( L10N_UNSUPPORTED_TARGET_IS_OPEN_ERROR_MESSAGE );
   return 0;
 }
+
+int 
+stumpless_entry_to_errorstream( const char *buffer, int buffer_len ){
```
goatshriek: With the above streamlining, this may not need to be a separate function as it will only be called once. If you do decide to keep it, change it to a private function (move to the `include/private/target.h` header) and rename it to something like `write_buffer_to_error_stream`.

src/target.c
```diff
@@ -108,61 +111,107 @@ stumpless_add_entry( struct stumpless_target *target,
...
   }
 
-  builder = format_entry( entry, target );
-  if( !builder ) {
-    return -1;
-  }
-
-  buffer = strbuilder_get_buffer( builder, &builder_length );
+  // entry was not formatted before
+  if( is_log_formatted == false){
```
goatshriek: Change this check to see if `builder` is `NULL` instead, to remove the need for a separate boolean flag.

include/stumpless/option.h
```diff
@@ -71,6 +71,15 @@
 #    define STUMPLESS_OPTION_NOWAIT (1<<4)
 #  endif
 
+/** Log the message to the stderr as well as
+ *  to the target
+*/
+# ifdef STUMPLESS_SYSLOG_H_COMPATIBLE
+#   define STUMPLESS_OPTION_PERROR LOG_PERROR
```
goatshriek, line 78: I did not realize this had not been defined yet, good catch! Please add an entry for `STUMPLESS_OPTION_PERROR` in `tools/check_headers/stumpless.yml` so that it will be picked up by the header checking tool.

goatshriek (changes requested):
Overall, it looks like the general implementation is here, well done! Here are a number of adjustments for you to make, which should help with the test coverage, which is currently too low.

Jk-bit:
New commit on my working branch, I have less experience working on test coverage and stuff. I have resolved the requested changes.

include/stumpless/target.h
```diff
@@ -950,6 +950,7 @@ vstumpless_add_message( struct stumpless_target *target,
                         const char *message,
                         va_list subs );
 
+
```
goatshriek: remove this empty line so that this file is not touched by this change

src/target.c
```diff
@@ -17,6 +17,8 @@
  */
 
 #include <stdarg.h>
+#include <stdio.h>
+#include <stdbool.h>
```
goatshriek: see the failed code analysis check - the `stdbool.h` header should be removed as it is unnecessary, and the `stumpless/error.h` header should be added for the use of `stumpless_get_error_stream`.

src/target.c
```diff
@@ -108,9 +111,26 @@ stumpless_add_entry( struct stumpless_target *target,
...
+	error_stream = stumpless_get_error_stream();
+  	if( fwrite( buffer, 1, builder_length, error_stream) != builder_length ){
+    	  return -1;
+	}
+  }
 
   // function targets are not formatted
-  if( target->type == STUMPLESS_FUNCTION_TARGET ) {
+  if( target->type == STUMPLESS_FUNCTION_TARGET ) {	
+    // Checking for the PERROR option
```
goatshriek: these lines should not be affected any more - remove the comment and make sure whitespacing is not changed on the conditional so that it is no longer in the diff

src/target.c
```diff
@@ -125,16 +145,18 @@ stumpless_add_entry( struct stumpless_target *target,
...
+  	if( !builder ) {
+    		return -1;
+  	}
+  	buffer = strbuilder_get_buffer( builder, &builder_length );
 
+  }  
   switch ( target->type ) {
 
-    case STUMPLESS_BUFFER_TARGET:
+    case STUMPLESS_BUFFER_TARGET:
```
goatshriek: remove the trailing whitespace here

src/target.c
```diff
@@ -729,3 +751,4 @@ unsupported_target_is_open( const struct stumpless_target *target ) {
   raise_target_unsupported( L10N_UNSUPPORTED_TARGET_IS_OPEN_ERROR_MESSAGE );
   return 0;
 }
+
```
goatshriek: remove this extra newline

goatshriek (changes requested):
Well done, a few more fine-tuning changes and this feature should be finished!

You are seeing a number of failures in the CI runs due to a combination poor design decisions made during the implementation of multithreading support and testing of unsupported targets. The retrieval of the options field for the target is reaching past the stack-allocated target structure in a number of tests, causing undefined behavior and sporadic failures. If you're curious about it, you can check out the target struct definition in `include/stumpless/target.h`, specifically the note at the end of it.

This isn't your problem though. I will create a patch for this and push it to the `latest` branch, after which I will re-test your finished changes and merge them in once I've confirmed that the issue resolves these failures.

goatshriek:
Struct issue has been corrected via #204, a separate branch (`test-fix`) is in work to fix the tests with undefined behavior.

goatshriek:
When possible, please merge in the recent changes in the `latest` branch and push the updates to confirm that the CI tests have been fixed.

PR #204, opened 2021-12-19
add mutexes to structure definitions

goatshriek:
In the initial implementation of thread safety, the mutex used to protect target, entry, element, and param types was stored at the end of the struct in an effort to avoid a dereference operation and simplify initialization.

Unfortunately, this leads to a mismatch in the memory required for a proper target structure and the size of the structure according to the compiler. This in turn can lead to undefined behavior when using a stack-allocated structure. While this isn't currently supported usage, it is planned for future releases, and is also used by some test suites (unsupported target types being one) which leads to instability.

This change adds a void pointer to the end of the structure instead which points at the mutex to use for synchronization of these types.

The need for this was originally encountered in #202.


goatshriek:
Merging despite low coverage to assist with #202.

PR #206, opened 2021-12-20
Cons spec

Jk-bit:
Runtime specifications for Console stream


Jk-bit:
Sorry for the extra commits seen, by mistake I didn't checkout from the "latest" branch.

include/stumpless/target.h
```diff
@@ -948,6 +949,49 @@ vstumpless_add_message( struct stumpless_target *target,
                         const char *message,
                         va_list subs );
 
+/**
+ *	Gets the current console stream where logs are written to. 	
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe. Atomic variables are used to store and
+ * retrieve the error stream.
```
goatshriek: change "error" to console

include/stumpless/target.h
```diff
@@ -948,6 +949,49 @@ vstumpless_add_message( struct stumpless_target *target,
...
+ *			the CONS option.
+ */
+FILE *
+stumpless_get_cons_stream( void );
+
+/**
+ * Sets the console stream to write logs to.
+ *
+ * This will be stdout by default, but can be set to any stream. If it is set
+ * to NULL then error messages will not be printed.
```
goatshriek: change "error" to console

include/stumpless/target.h
```diff
@@ -948,6 +949,49 @@ vstumpless_add_message( struct stumpless_target *target,
...
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe. Atomic variables are used to store and
+ * retrieve the error stream.
```
goatshriek: change "error" to console

include/stumpless/target.h
```diff
@@ -948,6 +949,49 @@ vstumpless_add_message( struct stumpless_target *target,
...
+ *
+ * **Async Signal Safety: AS-Safe**
+ * This function is safe to call from signal handlers.
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be ansynchronously
+ * cancelled.
+ *
+ * @param stream The stream to write logs to. If this is NULL then it will be
+ * ignored
```
goatshriek: The last sentence is unfortunately misleading in the error stream function. For this new one, change it to something more explicit such as "If this is NULL then messages intended for the console will be ignored." or something similar.

Feel free to change the corresponding language in the error stream function as well!

include/stumpless/target.h
```diff
@@ -948,6 +949,49 @@ vstumpless_add_message( struct stumpless_target *target,
...
+ * This function is thread safe. Atomic variables are used to store and
+ * retrieve the error stream.
+ *
+ * **Async Signal Safety: AS-Safe**
+ * This function is safe to call from signal handlers.
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be ansynchronously
+ * cancelled.
+ *
```
goatshriek, line 966: add an `@since v2.1.0` tag to annotate when this function was added

include/stumpless/target.h
```diff
@@ -948,6 +949,49 @@ vstumpless_add_message( struct stumpless_target *target,
...
```
goatshriek, line 990: add an `@since v2.1.0` tag here as well

src/windows/stumpless.def
```diff
@@ -158,4 +158,6 @@ EXPORTS
   stumpless_close_function_target               @143
   stumpless_open_function_target                @144
   stumpless_element_to_string                   @145
+  stumpless_get_cons_stream						@146
+  stumpless_set_cons_stream						@147
```
goatshriek: use spaces in this file instead of tabs - that is why they are not aligned in the github rendering above

src/target.c
```diff
@@ -53,6 +54,9 @@
 /* global static variables */
 static config_atomic_ptr_t current_target = config_atomic_ptr_initializer;
 static config_atomic_ptr_t default_target = config_atomic_ptr_initializer;
+static config_atomic_ptr_t cons_stream = config_atomic_ptr_initializer;
+static config_atomic_bool_t cons_stream_free = config_atomic_bool_true;
```
goatshriek, line 58: A different option here would be to use an atomic ptr to a `stream_target` struct, which would handle multithreaded access for you once created. However, it would need to be initialized when accessed for the first time, which might be more complicated than this approach. Ultimately it's your decision on which way to go, I just want to provide the option to you for consideration.

goatshriek (changes requested):
Excellent work, as usual! There are a few clean-ups for you to make, and then this can be merged.

The choice on the `stream_target` is yours: as I mentioned on the Gitter chat, this was not an option for the error stream, but is for the console one. However, as I reviewed this pull I see that it could be just as difficult to work with as the separate atomic variables, so I am not sure which would be the better approach.

src/target.c
Jk-bit: I guess creating a structure for the same will create unnecessary complications, defining the variables as it is might be a simple way to proceed

PR #208, opened 2021-12-22
updated perror option

Jk-bit:
Updated perror write

### DIFF
--- a/include/stumpless/error.h
+++ b/include/stumpless/error.h
@@ -324,6 +324,35 @@ stumpless_perror( const char *prefix );
 void
 stumpless_set_error_stream( FILE *stream );
 
+/**
+ *  @since v2.1.0
+ *
+ *  Logs the message to the error stream, ignore the message if the 
+ *  error stream is NULL
+ * 
+ * **Thread Safety: MT-Safe race:prefix**
+ * This function is thread safe, of course assuming that prefix is not changed
+ * by other threads during execution. A lock is used to coordinate accesses to
+ * the error stream.
+*
+ * **Async Signal Safety: AS-Unsafe lock**
+ * This function is not safe to call from signal handlers, as it uses a
+ * non-reentrant lock to synchronize access to the error stream.
+ *
+ * **Async Cancel Safety: AC-Unsafe lock**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, as the lock used to control access to the error stream may not
+ * be released after a cancellation.
+* 
+ * @param buffer message to be logged on the error stream.
+ * @param buff_len length of the message to be logged.
+ * 
+ * @return value -1 if the write is unsuccesful, 0 if successful
+ */
+int
+stumpless_log_to_error_stream( const char *buffer, int buff_len );
+
+
 #  ifdef __cplusplus
 }                               /* extern "C" */
 #  endif
--- a/src/error.c
+++ b/src/error.c
@@ -334,3 +334,21 @@ void
 raise_invalid_encoding( const char *message ) {
   raise_error( STUMPLESS_INVALID_ENCODING, message, 0, NULL );
 }
+
+int
+stumpless_log_to_error_stream( const char* buffer, int buff_len ) {
+  FILE *error_stream = stumpless_get_error_stream();
+  int ret;
+  bool locked;
+  if( error_stream == NULL ) {
+	return 0;
+  }
+  do{
+    locked = config_compare_exchange_bool( &error_stream_free, true, false );
+  } while( !locked );
+  ret = fwrite( buffer, 1, buff_len, error_stream );
+  if( ret != buff_len ) {
+    return -1;
+  } 
+  return 0;
+}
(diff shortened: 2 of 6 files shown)


include/stumpless/error.h
```diff
@@ -324,6 +324,35 @@ stumpless_perror( const char *prefix );
...
+ * cancelled, as the lock used to control access to the error stream may not
+ * be released after a cancellation.
+* 
+ * @param buffer message to be logged on the error stream.
+ * @param buff_len length of the message to be logged.
+ * 
+ * @return value -1 if the write is unsuccesful, 0 if successful
+ */
+int
+stumpless_log_to_error_stream( const char *buffer, int buff_len );
```
goatshriek, line 353: rename without the `stumpless_` prefix: `log_to_error_stream`

src/error.c
```diff
@@ -334,3 +334,21 @@ void
 raise_invalid_encoding( const char *message ) {
   raise_error( STUMPLESS_INVALID_ENCODING, message, 0, NULL );
 }
+
```
goatshriek, line 337: move the definition of the function into the private functions section of the file (the bottom portion)

src/windows/stumpless.def
```diff
@@ -160,4 +160,5 @@ EXPORTS
   stumpless_element_to_string                   @145
   stumpless_get_cons_stream                     @146
   stumpless_set_cons_stream                     @147
+  stumpless_log_to_error_stream                 @148
```
goatshriek, line 163: remove from .def file (no changes to be made)

tools/check_headers/stumpless.yml
```diff
@@ -452,3 +452,4 @@
 "winsock2_network_target_is_open": "private/config/have_winsock2.h"
 "stumpless_get_cons_stream" : "stumpless/target.h"
 "stumpless_set_cons_stream" : "stumpless/target.h"
+"stumpless_log_to_error_stream" : "stumpless/error.h"
```
goatshriek, line 455: move the function entry (with the new name and include file) into the `stumpless_private.yml` file instead

include/stumpless/error.h
```diff
@@ -324,6 +324,35 @@ stumpless_perror( const char *prefix );
 void
 stumpless_set_error_stream( FILE *stream );
 
+/**
```
goatshriek, line 327: move this function declaration into the `include/private/error.h` header file instead

goatshriek (changes requested):
This implementation looks fine, and should resolve the threading issue.

However, the new function should be made private, as it isn't something that needs to be exposed to users of the library. I've annotated the changes needed to do that below, once that is done I will merge this in to resolve the issue.

goatshriek:
Note that you will likely also need to do a fresh pull from the `latest` branch to resolve some conflicts from the latest update, which touched a fair amount of target module code.

src/target.c
```diff
@@ -126,10 +126,8 @@ stumpless_add_entry( struct stumpless_target *target,
     		return -1;
   	}
   	buffer = strbuilder_get_buffer( builder, &builder_length );
-	error_stream = stumpless_get_error_stream();
-  	if( fwrite( buffer, 1, builder_length, error_stream) != builder_length ){
-    	  return -1;
-	}
+	if( stumpless_log_to_error_stream( buffer, builder_length) == -1)
+      return -1;
```
igorkole, line 130: If perror write fails, the destination target is ignored.  Is this desired behavior?  After all, perror is just an option.

igorkole: Also, it returns without destroying the builder

src/error.c
```diff
@@ -334,3 +334,21 @@ void
...
+int
+stumpless_log_to_error_stream( const char* buffer, int buff_len ) {
+  FILE *error_stream = stumpless_get_error_stream();
+  int ret;
+  bool locked;
+  if( error_stream == NULL ) {
+	return 0;
```
igorkole, line 344: Why it returns 0?  Isn't it an error if there is no stream?

goatshriek: No, the error stream can be set to `NULL` to suppress error output entirely, so if it is `NULL` then it should simply be ignored.

src/target.c
goatshriek: Falling through here is a valid option, though an error should still be raised (from within the log to error stream function) so that the write failure can be detected by callers.

Good catch on the builder not being destroyed!

src/error.c
```diff
@@ -334,3 +334,21 @@ void
...
+  }
+  do{
+    locked = config_compare_exchange_bool( &error_stream_free, true, false );
+  } while( !locked );
+  ret = fwrite( buffer, 1, buff_len, error_stream );
+  if( ret != buff_len ) {
+    return -1;
```
goatshriek, line 351: add a call to `raise_stream_write_failure` here to capture the cause of the error

src/target.c
goatshriek: UPDATE: upon further inspection while working to fix this, I believe that @igorkole is correct on both fronts: the destination target shouldn't be ignored if the error stream writes fail. My apologies for not understanding the first time through. You'll be tagged as a coauthor in the commit that finally fixes this.

goatshriek:
This is being closed in favor of #235, as this pull has gone stale and attempts to reach the original author via the change requests on this PR and via Gitter have gone unanswered.

Issue #147, opened 2020-12-08
length checking of param names


goatshriek:
The syslog RFC ([RFC 5424](https://tools.ietf.org/html/rfc5424)) specifies a maximum length for many fields of a log message, including param names. Checking the length of supplied param name arguments will not only help users stay compliant with the standard, but will also allow for efficiencies in the implementation by eventually allowing for the use of fixed-length buffers instead of memory allocated from the heap.

As specified in the RFC, param names should be checked for their length and rejected if they are more than the allowed length (32 characters). Note that this length does not include the terminating NULL character from the C string representation.

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block. You are also free to take a different approach if you feel it is better - this is merely meant as a starting point.

There are a number of functions that allow parameter name to be supplied - the constructor for params, and the functions that allow new params to be added to elements and entries directly. Since this check will need to take place in several places, you will be best served by implementing it in a single function and then calling it wherever required.

Declare your new function (something like `validate_param_name_length`) in `include/private/validate.h`, and define it in `src/validate.c`. Similar validation functions have already been created for msgids and app names - use them as a reference as needed. After your new function has been defined, add calls to it wherever a param name is passed as a parameter in the library, which are the following functions:
 * `stumpless_new_param` (in `src/param.c`)
 * `stumpless_set_param_name` (also in `src/param.c`)
 * `stumpless_add_new_param` (in `src/element.c`)
 * `stumpless_add_new_param_to_entry` (in `src/entry.c`)

Note that you don't need to validate names passed to search functions like `stumpless_get_param_by_name` as the search will simply fail for non-compliant names.

Update the documentation for each of these functions to specify that the param name length is restricted to be 32 characters or less. The relevant documentation is in a doxygen comment format in the include files corresponding to the source files (`include/stumpless/param.h`, `include/stumpless/element.h`, and `include/stumpless/entry.h`). Add the note about this to the `@param` descriptions of the param name parameters.

[...]

PR #213, opened 2022-01-03
 length checking of param names

igorkole:
Do not allow parameter names to exceed the maximum length of 32, as per RFC5424.
Issue #147

### DIFF
--- a/include/private/validate.h
+++ b/include/private/validate.h
@@ -73,6 +73,17 @@ bool validate_app_name_length( const char* app_name);
 bool validate_param_name( const char* str);
 
 /**
+ * Checks that the passed in param name is of valid length.
+ *
+ * @param the param name(string).
+ *
+ * @return True if the param name is less than allowed length 
+ * (32 characters not including NULL terminating), otherwise
+ * it will return false and raise STUMPLESS_ARGUMENT_TOO_BIG error.
+ */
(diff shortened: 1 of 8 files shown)


goatshriek (changes requested):
Overall, this is excellent work! There are only two things that need to be adjusted before this can be merged.

First, please add tests for the `stumpless_add_new_param` and `stumpless_add_new_param_to_entry` functions. While the current implementation should already work, tests are needed to make sure there is no regression later.

Finally, please add entries for the new functions you have implemented to `tools/check_headers/stumpless_private.yml` so that header checks function properly.

igorkole:
> Finally, please add entries for the new functions you have implemented to `tools/check_headers/stumpless_private.yml` so that header checks function properly.

@goatshriek there are no validate_* functions in any of .yml files.  Should they all be added to stumpless_private.yml?

goatshriek:
You can if you'd like (they will need to go in eventually), but the only ones _needed_ are those you've implemented for this change.

goatshriek:
Well somewhat surprisingly, it looks like this has uncovered a missing header in `src/validate.c` itself! Looks like that is the last issue that needs to be resolved now.

igorkole:
I've added the missing header and excluded "validate_string_length" from .yml as it's internal to the validate.c module.

PR #233, opened 2022-09-10
Closes #78: Implement Option CONS.

### LINKED ISSUE #78: implement CONS option
Stumpless currently does not support the `STUMPLESS_OPTION_CONS` option despite providing it in its header files. This option is meant to mirror the `LOG_CONS` option from `syslog.h` in functionality, which causes messages to be written to the system console if there is a problem sending them to the target.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here and someone can help you get past the stumbling block. You are also free to take a different approach if you feel it is better - this is merely meant as a starting point.

First, have a look at the documentation for `stumpless_get_option`, `stumpless_set_option`, and `stumpless_unset_option` in `include/stumpless/target.h`. You will not need to modify any of these, but you will use them in your implementation so they are important to understand. Also be sure to understand how `stumpless_get_cons_stream` works, as you'll need it to retrieve the current console stream.

Next, check out `stumpless_add_entry` in `src/target.c`. This is the workhorse function for logging messages, and is a good place to put the logic to handle this option. Get the result from the calls to the sendto functions, and if the result code indicates failure (less than zero) and the option is set on the target, then print the formatted message to the console.

Don't forget to document the option after you implement it in `include/stumpless/option.h`.

Once you have implemented this, be sure to add a test for the functionality. Test code already exists to make sure that setting and unsetting the option behaves as expected - look for `TEST( SetOption, Pid )` in `test/function/target.cpp`. Use this test as a starting point to create a new test that logs a message before and after setting the option. To cause a failure, you can use a stream target with a read only stream so that attempts to write to it fail. You can see an example of this strategy in the unit test suite for stream targets - look for `TEST( StreamTargetWriteTest, ReadOnlyStream )` in `test/function/target/stream.cpp`. To capture output, set the console stream to your own custom stream using `stumpless_set_cons_stream`.

matthew-macgregor:
# Overview

Implements the `STUMPLESS_OPTION_CONS`, which will (see caveats) write to the console if sending to the configured target responds with a failure return status.

## Details & Caveats

- Must be configured on a per-target basis.
- Considered a last resort. Any error caused by the console write will be ignored. The error from the original target send will be preserved.
- Will only write in the case that the target has identified an error (returned a result of < 0). There are failure cases for individual targets/operating systems that might bypass this (both network and file targets have caveats, POSIX signals, etc.).

Discussion is here: https://github.com/goatshriek/stumpless/issues/78

## Testing

- Automated "smoke tests" for the stream target.
- Tested on Ubuntu 22.04 and macOS 12.5.1 with the network and file targets.
- TODO: Testing Windows, other UNIX/BSD, additional targets.

## Questions & Other notes

- Lots of duplication in my tests. I stuck with this style since it's usually best not to overload a single test case (and not sure what the right patterns are for reuse). What do you prefer?
- Future idea: test harness (separate repo?) that builds and runs a set of functional tests against "live" targets. Inspired by the [SIGPIPE MRE](https://github.com/matthew-macgregor/stumpless-network-sigpipe-example) I put together, but actually testing something.
- Thanks for the cool project and great getting started instructions! I had fun working on this and learned some new things. I don't get to use C or C++ at work, so it's nice to have the opportunity to contribute in my free time. Gave me reason to crack open my Linux System Programming book, too!
- Would love any and all feedback.


src/target.c
```diff
@@ -355,22 +368,35 @@ stumpless_get_default_target( void ) {
 }
 
 int
-stumpless_get_option( const struct stumpless_target *target, int option ) {
+unchecked_get_option( const struct stumpless_target *target, int option ) {
   int options;
 
-  if( !target ) {
-    raise_argument_empty( L10N_NULL_ARG_ERROR_MESSAGE( "target" ) );
+  if ( !target ) {
```
goatshriek: I believe this check is why your coverage has dropped a little - you can remove it. The `unchecked_` family of functions explicitly put the burden of checks on the caller, and in the usages of this function `target` has already been checked for NULL earlier.

I'll add some documentation on this naming convention and expectation as well as some others that up until this point have just existed in my head. These will go in `docs/style.md` shortly.

Edit: See #236 for the documentation update to describe naming conventions.

test/function/target.cpp
```diff
@@ -998,6 +998,30 @@ namespace {
...
+  TEST( SetOption, Cons ) {
+    struct stumpless_target *target;
+    struct stumpless_target *target_result;
+    char buffer[100];
+    int option;
+
+    target = stumpless_open_buffer_target( "test target",
+                                           buffer,
+                                           sizeof( buffer ) );
+    ASSERT_TRUE( target != NULL );
```
goatshriek: Consider changing this to `ASSERT_NOT_NULL`. This isn't completely consistently used across the test base yet, but I am pushing to move it in that direction for new tests.

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+    char buffer[300];
+    // Ensure that the buffer is cleared
+    memset( buffer, 0, 300 );
+
+    // Create the temporary file, then open as read only
+    ro_stream = fopen( ro_filename, "w+" );
+    fclose( ro_stream );
+
+    ro_stream = fopen( ro_filename, "r" );
+    ASSERT_TRUE( ro_stream != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here as well

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+
+    // Open fallback to capture output
+    cons_stream = fopen( cons_filename, "w+" );
+    ASSERT_TRUE( cons_stream != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here as well

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+
+    // Use this to capture the fallback output
+    stumpless_set_cons_stream( cons_stream );
+    ASSERT_TRUE( cons_stream == stumpless_get_cons_stream( ) );
+
+    target = stumpless_open_stream_target( ro_filename, ro_stream );
+    ASSERT_TRUE( target != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here as well

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+
+    basic_entry = stumpless_new_entry( STUMPLESS_FACILITY_USER,
+                                       STUMPLESS_SEVERITY_INFO,
+                                       "stumpless-unit-test",
+                                       "basic-entry",
+                                       "STUMPLESS_OPTION_CONS::01" );
+    ASSERT_TRUE( basic_entry != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here as well

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+    result = stumpless_add_entry( target, basic_entry );
+
+    // Expect the write to fail and underlying error preserved
+    EXPECT_LT( result, 0 );
+    EXPECT_ERROR_ID_EQ( STUMPLESS_STREAM_WRITE_FAILURE );
+
+    // Confirm that we didn't write to the console stream
+    ASSERT_TRUE( fgets( buffer, 300, cons_stream ) == NULL );
```
goatshriek: use `ASSERT_NULL` here instead

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+
+    FILE *ro_stream;
+    FILE *cons_stream;
+
+    // Create the temporary file, then open as read only
+    ro_stream = fopen( ro_filename, "w+" );
+    fclose( ro_stream );
+
+    ro_stream = fopen( ro_filename, "r" );
+    ASSERT_TRUE( ro_stream != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here as well

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+
+    cons_stream = fopen( cons_filename, "w+" );
+    ASSERT_TRUE( cons_stream != NULL );
+
+    // Use this to capture the fallback output
+    stumpless_set_cons_stream( cons_stream );
+    ASSERT_TRUE( cons_stream == stumpless_get_cons_stream( ) );
+
+    target = stumpless_open_stream_target( ro_filename, ro_stream );
+    ASSERT_TRUE( target != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here as well

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+    // fallback if logging to the target failed
+    target_result = stumpless_set_option( target, STUMPLESS_OPTION_CONS );
+    EXPECT_EQ( target_result, target );
+
+    basic_entry = stumpless_new_entry( STUMPLESS_FACILITY_USER,
+                                       STUMPLESS_SEVERITY_INFO,
+                                       "stumpless-unit-test",
+                                       "basic-entry",
+                                       message );
+    ASSERT_TRUE( basic_entry != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here as well

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+    // Write should fail because the target is read only!
+    EXPECT_LT( result, 0 );
+    EXPECT_ERROR_ID_EQ( STUMPLESS_STREAM_WRITE_FAILURE );
+
+    // Ensure that the buffer is cleared
+    memset( buffer, 0, 300 );
+    rewind( cons_stream );
+
+    // Confirm that the cons_stream is empty
+    ASSERT_TRUE( fgets( buffer, 300, cons_stream ) != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here as well

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+
+    // Write should fail because the target is read only!
+    EXPECT_LT( result, 0 );
+    EXPECT_ERROR_ID_EQ( STUMPLESS_STREAM_WRITE_FAILURE );
+
+    // Ensure that the buffer is cleared
+    memset( buffer, 0, 300 );
+    rewind( cons_stream );
+
+    // Confirm that the cons_stream is empty
```
goatshriek: seems like we're confirming that the stream is NOT empty here?

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+    int result;
+
+    FILE *ro_stream;
+
+    // Create the temporary file, then open as read only
+    ro_stream = fopen( ro_filename, "w+" );
+    fclose( ro_stream );
+
+    ro_stream = fopen( ro_filename, "r" );
+    ASSERT_TRUE( ro_stream != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here as well

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+
+    // Use this to capture the fallback output
+    stumpless_set_cons_stream( NULL );
+
+    target = stumpless_open_stream_target( ro_filename, ro_stream );
+    ASSERT_TRUE( target != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here as well

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+    // fallback if logging to the target failed
+    target_result = stumpless_set_option( target, STUMPLESS_OPTION_CONS );
+    EXPECT_EQ( target_result, target );
+
+    basic_entry = stumpless_new_entry( STUMPLESS_FACILITY_USER,
+                                       STUMPLESS_SEVERITY_INFO,
+                                       "stumpless-unit-test",
+                                       "basic-entry",
+                                       message );
+    ASSERT_TRUE( basic_entry != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here as well

test/function/target.cpp
```diff
@@ -1055,6 +1080,197 @@ namespace {
...
+    result = stumpless_add_entry( target, basic_entry );
+
+    // Write should fail because the target is read only!
+    EXPECT_LT( result, 0 );
+    EXPECT_ERROR_ID_EQ( STUMPLESS_STREAM_WRITE_FAILURE );
+
+    // Confirm that our custom console is null (but did not segfault)
+    ASSERT_TRUE( NULL == stumpless_get_cons_stream( ) );
```
goatshriek: use `ASSERT_NULL` here as well

goatshriek (changes requested):
Overall, this is excellent work! Thank you very much for putting it together, particularly for the detailed testing (both in this pull and the SIGPIPE issue). Most of these changes are just minor style changes. Removing the redundant NULL check should bring your coverage up to an increase and remove the one failing CI check. I'll merge this once I see those changes come through.

Don't worry too much about the duplication in your tests. I tend to relax the anti-duplication standards for these tests, and only go back and refactor them when Sonarcloud highlights an issue.

A live test harness simulation would be a huge improvement, and is something that I'll need to put some time into once I start adding more features to the network targets, like support for compression and RFC 5425 (which I do intend to do eventually). I haven't had time to dive into your MRE yet, but that's next in my queue after the WEL work that is closing up. I'm sure it will have some useful tidbits that I can reuse to make the network testing more robust.

Again, your contribution is much appreciated! A main goal of this project is to serve as a learning tool, so any suggestions you have on documentation or other things that would lower the barrier to entry to newcomers would be welcome.

matthew-macgregor:
Thank you for the detailed review! I'll make these changes ~soon~.

test/function/target.cpp
matthew-macgregor: While I am at it I'll make this change for the rest of the cases in `test/function/target.cpp` which are using `ASSERT_TRUE` to check for `NULL`.

matthew-macgregor:
@goatshriek, took me a bit longer to get back to this than I expected, but I've updated the PR and pushed my changes!

Issue #237, opened 2022-09-15
facility and severity from string functions


goatshriek:
Stumpless uses enumerations for some of its settings, most notably the severity and facility codes used in syslog entries. While functions exist to get a string representation of these values, there are not functions that operate in the reverse; that is, are able to return an enumeration type given a string. Such functions would be useful in some scenarios, for example in the command line parsing performed by the [CLI logger](https://github.com/goatshriek/stumpless-logger) being built as a wrapper of stumpless.

This task is to implement this functionality for the facility and severity enumerations. These functions should accept string names of the enumeration in question and return a matching enumeration value. At a minimum the string returned by `stumpless_get_facility_string` or `stumpless_get_severity_string` should be converted. A more ambitious implementation will also be case insensitive and/or also accept just the name of the facility/enumeration, but if you don't feel up to it then these features can be done as a separate change later.

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block. You are also free to take a different approach if you feel it is better - this is merely meant as a starting point.

First, check out the implementations of the `stumpless_get_facility_string` and `stumpless_get_severity_string` functions (in `src/facility.c` and `src/severity.c`). These demonstrate how to use the `FOREACH` macros with other macro functions to enumerate over the possible values. These functions rely on tables of strings to go from an index to a string value.

Your implementation may simply loop through this array, checking for string matches and returning if one is found. Or, you may choose to do something else with the FOREACH macro construct if desired. You can also take an entirely different approach if you'd like. The only thing to avoid is hardcoding the values from the enumeration - this must be done dynamically.

Once these functions are implemented, add tests for them to the appropriate test suites. These should go in the test suites `test/function/facility.cpp` and `test/function/severity.cpp`.

PR #239, opened 2022-09-20
missing tests

salustiana:
Implemented functions returning facility and severity enums from their corresponding strings.

#237 

First time contributing, let me know if this is in the right direction, or if I'm missing anything. If this seems alright I'll write the tests for these two functions.

### DIFF
--- a/include/stumpless/facility.h
+++ b/include/stumpless/facility.h
@@ -372,6 +372,31 @@ STUMPLESS_PUBLIC_FUNCTION
 const char *
 stumpless_get_facility_string( enum stumpless_facility facility );
 
+  /**
+ * Gets the enum value corresponding to the given facility string.
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe.
+ *
+ * **Async Signal Safety: AS-Safe**
+ * This function is safe to call from signal handlers.
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled.
+ *
+ * @param facility_string The facility name to get the enum from.
+ *
+ * @return The enum integer corresponding to the given facility or -1 if
+ * the string is not a valid facility name.
+ *
+ * TODO: is this correct?
+ * @since release v2.1.0.
+ */
+STUMPLESS_PUBLIC_FUNCTION
+enum stumpless_facility
+stumpless_get_facility_enum( const char *facility_string );
+
 #  ifdef __cplusplus
 } /* extern "C" */
 #  endif
--- a/include/stumpless/facility.h
+++ b/include/stumpless/facility.h
@@ -362,11 +362,11 @@ enum stumpless_facility {
  * This function is safe to call from threads that may be asynchronously
  * cancelled.
  *
+ * @since release v2.1.0.
+ *
  * @param facility The facility to get the string from.
  *
  * @return The string representation of the given facility.
- *
- * @since release v2.1.0.
  */
 STUMPLESS_PUBLIC_FUNCTION
 const char *
@@ -385,13 +385,12 @@ stumpless_get_facility_string( enum stumpless_facility facility );
  * This function is safe to call from threads that may be asynchronously
  * cancelled.
  *
+ * @since release v2.1.0.
+ *
  * @param facility_string The facility name to get the enum from.
  *
  * @return The enum integer corresponding to the given facility or -1 if
  * the string is not a valid facility name.
- *
- * TODO: is this correct?
- * @since release v2.1.0.
  */
 STUMPLESS_PUBLIC_FUNCTION
 enum stumpless_facility
--- a/include/stumpless/severity.h
+++ b/include/stumpless/severity.h
@@ -230,13 +230,12 @@ stumpless_get_severity_string( enum stumpless_severity severity );
  * This function is safe to call from threads that may be asynchronously
  * cancelled.
  *
+ * @since release v2.1.0.
+ *
  * @param severity_string The severity name to get the enum from.
  *
  * @return The enum integer corresponding to the given severity or -1 if
  * the string is not a valid severity name.
- *
- * TODO: is this correct?
- * @since release v2.1.0.
  */
 STUMPLESS_PUBLIC_FUNCTION
 enum stumpless_severity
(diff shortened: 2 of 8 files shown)


goatshriek:
This initial implementation looks fine! Once you have some tests with full coverage, let me know and I will do a full review and merge this in.

salustiana:
I've written the tests and I'm _hoping_ they'll work, since I'm running into these errors while trying to make the check target:
```
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libgtest.a(gtest-all.cc.o): undefined reference to symbol 'pthread_getspecific@@GLIBC_2.2.5'
/usr/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
```
and I couldn't figure them out.

goatshriek:
Hmm, that build failure appears to be with Google Test, though I haven't encountered it before on the current latest branch. You might be able to resolve it locally by passing `-pthread` as a compiler flag, but it should be resolved in the project itself if it's something that you're running into. Can you share some more details on the system and compiler that you're working with? For some reason none of the CI builds seem to be having the same problem.

salustiana:
> Hmm, that build failure appears to be with Google Test, though I haven't encountered it before on the current latest branch. You might be able to resolve it locally by passing `-pthread` as a compiler flag, but it should be resolved in the project itself if it's something that you're running into. Can you share some more details on the system and compiler that you're working with? For some reason none of the CI builds seem to be having the same problem.

sure, I was building on a debian distro with gcc. I'm not on it right now so I can't get a lot of info. But let me know what details are of interest and next time I'm on that system I'll let you know.

include/stumpless/facility.h
```diff
@@ -372,6 +372,31 @@ STUMPLESS_PUBLIC_FUNCTION
...
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled.
+ *
+ * @param facility_string The facility name to get the enum from.
+ *
+ * @return The enum integer corresponding to the given facility or -1 if
+ * the string is not a valid facility name.
+ *
+ * TODO: is this correct?
```
goatshriek: yes, this is the correct usage of since, and is the version this will be released in. However, the `@since` tags are above the param and return tags (and after the POSIX safety documentation) in the rest of the functions, so please move this one there as well.

include/stumpless/severity.h
```diff
@@ -217,6 +217,31 @@ STUMPLESS_PUBLIC_FUNCTION
...
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled.
+ *
+ * @param severity_string The severity name to get the enum from.
+ *
+ * @return The enum integer corresponding to the given severity or -1 if
+ * the string is not a valid severity name.
+ *
+ * TODO: is this correct?
```
goatshriek: Also correct, but move above the first `@param` here as well.

src/windows/stumpless.def
```diff
@@ -187,3 +187,7 @@ EXPORTS
   stumplog_str                                  @172
   stumplog_trace_str                            @173
   stumpless_get_facility_string                 @174
+
+; added in v2.1.0
```
goatshriek: you can remove the blank space and comment - this entire block of functions is new to v2.1.0

src/facility.c
```diff
@@ -33,6 +34,15 @@ stumpless_get_facility_string( enum stumpless_facility facility ) {
   }
   return "NO_SUCH_FACILITY";
 }
+
+enum stumpless_facility
+stumpless_get_facility_enum( const char *facility_string ) {
+  for (int i = 0; i < sizeof facility_enum_to_string / sizeof facility_enum_to_string[0]; i++)
+    if (strcmp(facility_string, facility_enum_to_string[i]) == 0)
+      return i * sizeof facility_enum_to_string[0];
```
goatshriek: This multiplication is incorrect, but I suspect it appears to work because the size of a pointer happens to be 8 on all of the current builds as they are 64-bit. I'll make a note to add in a CI build with different sizes to catch potential issues like this in the future.

In the meantime, please change this to `i << 3` , as is used in other facility value creations (see `include/stumpless/facility.h`).

goatshriek (changes requested):

include/stumpless/facility.h
salustiana: I copied the format from the functions above, but you're right: in all previous functions @since is above parameters. I'll take the liberty to change those functions as well, then.

src/facility.c
salustiana: yes, it also seemed strange to me, but I replicated the behavior from the "stumpless_get_facility_string" function above, which is defined:
```
 const char *
stumpless_get_facility_string( enum stumpless_facility facility ) {
  size_t facility_upper_bound = sizeof facility_enum_to_string;
  if ( !facility_is_invalid(facility) ) {
    return facility_enum_to_string[facility/ sizeof facility_enum_to_string[0]];
  }
  return "NO_SUCH_FACILITY";
}
```
the return is doing the same operation in reverse. I'll change that function as well, since I'm pretty sure it has the same error, but please let me know if I'm missing something and I'll revert it.

goatshriek:
Overall the pull request looks great, just a few minor changes above for you to make. I've opened #241 to address the build problems you're facing, along with some thoughts on what to try and what information might help reproduce it on my end.

Thanks so much for putting this together and persevering through build issues! The project will be much better off having the new features and build bugs worked out.

include/stumpless/facility.h
goatshriek: Augh, sorry about that, I seem to have missed that when I reviewed the change that introduced those. Thank you for fixing them as well!

src/facility.c
goatshriek: You're absolutely right, that function has the same issue as well! A different size architecture build is much needed, it seems. Thank you for finding/correcting those issues as well.

salustiana: no problem at all, thanks for the detailed feedback.

salustiana:
Thank you for helping me make my first contribution to open source!

goatshriek:
Absolutely, thank you for taking the time! I do try to highlight help that folks provide; if you'd like a shoutout on Twitter, please let me know your handle here or in a private message and I'll publish something there as well.

salustiana:
I'm not on twitter so don't worry about it. I appreciate it anyways!

PR #252, opened 2022-10-08
Closes #244: Network target protocol accessor functions

### LINKED ISSUE #244: network target protocol accessor functions
Network targets are created with specific network and transport protocols, for example IPv6 (network) over TCP (transport). However, there is not currently a public-facing function to retrieve this information about a target after it has been created. Your task is to add these new functions.

## General Approach

First, read [this](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions) section of the development documentation on adding new functions. Once you understand how this process works, start defining the two new functions. They should have the following signatures:

```c
STUMPLESS_PUBLIC_FUNCTION
enum stumpless_network_protocol
stumpless_get_network_protocol( const struct stumpless_target *target );

STUMPLESS_PUBLIC_FUNCTION
enum stumpless_transport_protocol
stumpless_get_transport_protocol( const struct stumpless_target *target );
```

Your new functions will be declared in `include/stumpless/target/network.h` and defined in `src/target/network.c`. Be sure that you handle the possibility of a non-network target being provided (other functions in the `network.c` module show how to handle this).

Finally, once you've implemented the new functions, be sure you add tests for them! These should go in `test/function/target/network.cpp` and exercise all possibilities (normal behavior, NULL argument, incompatible target, anything else you can think of).

Harry-Ramsey:
### Overview
Implements the network target protocol accessor functions to allow retrieving of network protocol and transport protocol stored in `network_targets`.

### Details & Caveats

In order to compile without errors/warnings we cannot return `NULL` and instead have to cast `-1` to the valid return enum. A test case also has been renamed to reflect its true intent and to not clash with the new test cases being implemented.

### Testing

Add network target protocol accessor unit tests.
TODO: Testing on Windows

### DIFF
--- a/include/stumpless/target/network.h
+++ b/include/stumpless/target/network.h
@@ -669,6 +669,55 @@ struct stumpless_target *
 stumpless_set_udp_max_message_size( struct stumpless_target *target,
                                     size_t max_msg_size );
 
+/**
+ * Gets the network protocol of a network target.
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe. A mutex is used to coordinate the read of the
+ * target with other accesses and modifications.
+ *
+ * **Async Signal Safety: AS-Unsafe lock heap**
+ * This function is not safe to call from signal handlers due to the use of a
+ * non-reentrant lock to coordinate access.
+ *
+ * **Async Cancel Safety: AC-Unsafe lock heap**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, due to the use of a lock that could be left locked.
+ *
+ * @param target The target to get the network protocol from.
+ *
+ * @return The network protocol of the network target.
+ * In the event of an error, -1 is returned and an error code is 
+ * set appropriately.
+ */
+STUMPLESS_PUBLIC_FUNCTION
+enum stumpless_network_protocol
+stumpless_get_network_protocol( const struct stumpless_target *target );
+
+/**
+ * Gets the transport protocol of a network target.
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe. A mutex is used to coordinate the read of the
+ * target with other accesses and modifications.
+ *
+ * **Async Signal Safety: AS-Unsafe lock heap**
+ * This function is not safe to call from signal handlers due to the use of a
+ * non-reentrant lock to coordinate access.
+ *
+ * **Async Cancel Safety: AC-Unsafe lock heap**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, due to the use of a lock that could be left locked.
+ *
+ * @param target The target to get the transport protocol from.
+ *
+ * @return The transport protocol of the network target. In the event of 
+ * an error, -1 is returned and an error code is set appropriately.
+ */
+STUMPLESS_PUBLIC_FUNCTION
+enum stumpless_transport_protocol
+stumpless_get_transport_protocol( const struct stumpless_target *target );                     
+
 #  ifdef __cplusplus
 }                               /* extern "C" */
 #  endif
--- a/src/windows/stumpless.def
+++ b/src/windows/stumpless.def
@@ -189,3 +189,5 @@ EXPORTS
   stumpless_get_facility_string                 @174
   stumpless_get_facility_enum                   @175
   stumpless_get_severity_enum                   @176
+  stumpless_get_network_protocol                @177
+  stumpless_get_transport_protocol              @178
--- a/test/function/target/network.cpp
+++ b/test/function/target/network.cpp
@@ -112,7 +112,7 @@ namespace {
     EXPECT_ERROR_ID_EQ( STUMPLESS_ARGUMENT_EMPTY );
   }
 
-  TEST( NetworkTargetGetTransportProtocol, BadTargetType ) {
+  TEST( NetworkTargetGetTransportPort, BadTargetType ) {
     const char *result;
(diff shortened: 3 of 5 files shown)


Harry-Ramsey:
I shall sort these issues out and get the commit signed in the morning.

include/stumpless/target/network.h
```diff
@@ -669,6 +669,54 @@ struct stumpless_target *
...
+ * non-reentrant lock to coordinate access.
+ *
+ * **Async Cancel Safety: AC-Unsafe lock heap**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, due to the use of a lock that could be left locked.
+ *
+ * @param target The target to get the transport protocol from.
+ *
+ * @return The transport protocol of the network target. In the event of 
+ * an error, NULL is returned and an error code is set appropriately.
```
goatshriek: update this to be -1 in the case of an error instead of NULL.

include/stumpless/target/network.h
```diff
@@ -669,6 +669,54 @@ struct stumpless_target *
...
+ * non-reentrant lock to coordinate access.
+ *
+ * **Async Cancel Safety: AC-Unsafe lock heap**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, due to the use of a lock that could be left locked.
+ *
+ * @param target The target to get the network protocol from.
+ *
+ * @return The network protocol of the network target.
+ * In the event of an error, NULL is returned and an error code is
```
goatshriek: update this to be -1 in the case of an error instead of NULL.

include/stumpless/target/network.h
```diff
@@ -669,6 +669,54 @@ struct stumpless_target *
...
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, due to the use of a lock that could be left locked.
+ *
+ * @param target The target to get the network protocol from.
+ *
+ * @return The network protocol of the network target.
+ * In the event of an error, NULL is returned and an error code is 
+ * set appropriately.
+ */
+enum stumpless_network_protocol
```
goatshriek, line 694: The `STUMPLESS_PUBLIC_FUNCTION` designator is missing here, which is causing the build errors you are seeing on Linux. On Windows, the functions need to be added to `src/windows/stumpless.def` to make them publicly visible.

test/function/target/network.cpp
```diff
@@ -521,4 +521,86 @@ namespace {
...
+  TEST ( NetworkTargetGetTransportProtocol, BadTargetType ) {
+    enum stumpless_transport_protocol result;
+    const struct stumpless_target *target;
+    const struct stumpless_error *error;
+    char buffer[100];
+
+    target = stumpless_open_buffer_target( "not-a-udp-target",
+                                           buffer,
+                                           sizeof( buffer ) );
+    ASSERT_TRUE( target != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here instead to improve readability

test/function/target/network.cpp
```diff
@@ -521,4 +521,86 @@ namespace {
...
+  TEST ( NetworkTargetGetNetworkProtocol, BadTargetType ) {
+    enum stumpless_network_protocol result;
+    const struct stumpless_target *target;
+    const struct stumpless_error *error;
+    char buffer[100];
+
+    target = stumpless_open_buffer_target( "not-a-udp-target",
+                                           buffer,
+                                           sizeof( buffer ) );
+    ASSERT_TRUE( target != NULL );
```
goatshriek: use `ASSERT_NOT_NULL` here instead to improve readability

goatshriek (changes requested):
This looks good overall! I went ahead with a review to give you some hints on what is still failing. Once these updates are made, this should be good to merge.

Harry-Ramsey:
I believe I have made all those changes now.

src/windows/stumpless.def
```diff
@@ -189,3 +189,5 @@ EXPORTS
   stumpless_get_facility_string                 @174
   stumpless_get_facility_enum                   @175
   stumpless_get_severity_enum                   @176
+  stumpless_get_network_protocol                @244
+  stumpless_get_transport_protocol              @244
```
goatshriek: the ordinals for these should be sequential: 177 and 178

goatshriek (changes requested):

src/windows/stumpless.def
Harry-Ramsey: Resolved :)

goatshriek:
Ahh, I forgot about the case where network targets are disabled, which is why the all features disabled build is failing on Windows. You'll need to add definitions to the `src/config/network_unsupported.c` file for the two new functions that simply raises a target unsupported error and returns -1, similar to the others in the file. That should clear everything up. Sorry about that!

Harry-Ramsey:
> Ahh, I forgot about the case where network targets are disabled, which is why the all features disabled build is failing on Windows. You'll need to add definitions to the `src/config/network_unsupported.c` file for the two new functions that simply raises a target unsupported error and returns -1, similar to the others in the file. That should clear everything up. Sorry about that!

This should hopefully be the last thing now and it should pass all test cases 👍

goatshriek:
Looks like you're in the home stretch - just make sure those return -1 instead of NULL and fix the signatures to match, and you should be all set.

Harry-Ramsey:
Ah, I forgot that. Have updated `NULL` to `-1`.

Issue #256, opened 2022-10-10
fixed buffer for element names


goatshriek:
According to RFC 5424, the name of a structured data element may be anywhere between 1 and 32 printable ASCII characters. This means that a fixed buffer could be used to hold the name of an element in the structure, reducing the need for dynamic memory allocation calls and pointer dereferences. This task is to convert the name of the element in the struct from a pointer to a character buffer.

## General Approach
There are a few details left out of the following outline, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, review the definition of the element structure in `include/stumpless/element.h`. The name field will need to be changed to a character buffer of size 33 (to account for the terminating NULL character). You can look at the app_name and/or msgid fields of the entry structure definition in `include/stumpless/entry.h` as an example - these are already implemented using fixed buffers.

Once you are ready, change the definition of the field to be a character buffer of fixed length. Add a `#define` for the maximum length of the string (32) for other code to use if needed, similar to app_name and msgid do.

Finally, you'll need to fix the internal code that uses this field as a pointer. This will of course include the get/set methods for an element name as well as the constructor and destructor, but may include others as well. The compiler and test suite should help you find most of these places and update them.

PR #258, opened 2022-10-11
Use fixed buffer for element names (#256)

adsr:
Per #256, for hacktoberfest

### DIFF
--- a/src/element.c
+++ b/src/element.c
@@ -399,10 +399,8 @@ stumpless_new_element( const char *name ) {
     goto fail;
   }
 
-  element->name = copy_cstring_with_length( name, &( element->name_length ) );
-  if( !element->name ) {
-    goto fail_name;
-  }
+  snprintf( element->name, sizeof(element->name), "%s", name );
+  element->name_length = strlen( element->name );
 
   element->params = NULL;
   element->param_count = 0;
@@ -418,11 +416,6 @@ stumpless_new_element( const char *name ) {
   return element;
 
 fail_mutex:
-  free_mem( element->name );
-
-fail_name:
-  free_mem( element );
-
 fail:
   return NULL;
 }
@@ -430,10 +423,6 @@ fail:
 struct stumpless_element *
 stumpless_set_element_name( struct stumpless_element *element,
                             const char *name ) {
-  char *new_name;
-  size_t new_size;
-  const char *old_name;
-
   VALIDATE_ARG_NOT_NULL( element );
   VALIDATE_ARG_NOT_NULL( name );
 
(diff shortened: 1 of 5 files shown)


src/element.c
```diff
@@ -399,10 +399,8 @@ stumpless_new_element( const char *name ) {
     goto fail;
   }
 
-  element->name = copy_cstring_with_length( name, &( element->name_length ) );
-  if( !element->name ) {
-    goto fail_name;
-  }
+  snprintf( element->name, sizeof(element->name), "%s", name );
```
goatshriek: Refrain from using format-string based functions such as `snprintf` unless absolutely necessary. In this case, you can gather the length of the string first and then use a memcpy operation to move the string over.

src/element.c
```diff
@@ -418,22 +416,13 @@ stumpless_new_element( const char *name ) {
   return element;
 
 fail_mutex:
```
goatshriek, line 417: remove this label now that it is empty

src/element.c
```diff
@@ -418,22 +416,13 @@ stumpless_new_element( const char *name ) {
...
-  free_mem( element->name );
-
-fail_name:
-  free_mem( element );
```
goatshriek, line 424: the element will still need to be cleaned up in the event of a failure, so this error branch cannot be removed.

src/element.c
```diff
@@ -442,19 +431,11 @@ stumpless_set_element_name( struct stumpless_element *element,
...
-  if( !new_name ) {
-    goto fail;
-  }
-
-
   lock_element( element );
-  old_name = element->name;
-  element->name = new_name;
-  element->name_length = new_size;
+  snprintf( element->name, sizeof(element->name), "%s", name );
```
goatshriek: Again, remove the use of `snprintf` here as well. Use the definition of `stumpless_set_entry_app_name` as a template for what to do here, as the `strlen` can be done outside of the critical section to improve performance.

goatshriek (changes requested):
Thank you for putting this together, it is mostly there! There are a few changes noted above for you to work on. Also, run the header check if you can before pushing your updates to try to avoid another static analysis failure.

src/element.c
adsr: I left this label in because we need to free the element in this case as you mentioned in your comment below.

src/element.c
adsr: Ack

adsr (commented):
Thanks @goatshriek

goatshriek:
Everything looks good to go, you just need to remove the include that is no longer needed as flagged by the header check.

It looks like the coverage dropped slightly because this change exposed some missing explicit tests for the copy element functionality. I'll create a follow up issue to deal with that separately.

goatshriek (changes requested):
Remove the private/strhelper.h include from src/element.c to pass the header check.

adsr:
Thanks again @goatshriek. (I see now that you have to quote the glob path you pass to the check_headers script. I mistakenly assumed the intention was shell expansion.)

goatshriek (approved):

PR #260, opened 2022-10-11
avoid SIGPIPE termination

### LINKED ISSUE #231: Handle SIGPIPE for Network Target
### TL;DR

The network target will terminate the program on POSIX systems if the remote closes the connection. I think the simplest solution is to pass the `MSG_NOSIGNAL` flag to the `send()` function. This should prevent the signal and is scope to just that call.

Might be a very minor change (a few lines?), but this proposed solution needs to be validated.

Other approaches are listed below.

### Details

**Is your feature request related to a problem? Please describe.**

On POSIX systems, a broken pipe or closed socket will result in the SIGPIPE signal being raised, and the default behavior of the signal is to terminate the process. The stumpless TCP network target (among others?) doesn't handle SIGPIPE, which will kill the program if raised.

- Stumpless version: v2.0.0
- Platform: Any POSIX compatible OS? Observed on Ubuntu Linux 22.04 (5.15.0-47-generic).

I put together an MRE: https://github.com/matthew-macgregor/stumpless-network-sigpipe-example

**Workaround**

In the short term, the programmer should be aware of this behavior when using the network target, and can either `signal( SIGPIPE, SIG_IGN )` to ignore the signal, or use the more advanced signal handling functions to temporarily block then retrieve signals if more control is needed. Using a global signal handler is discouraged, in part because we can't be sure that stumpless is in a consistent state from inside the handler (and because there are better solutions).

**Describe the solution you'd like**

Since the underlying function `sys_socket_sendto_target` is using the `send()` function, `MSG_NOSIGNAL` should be passed in the flags argument.

This should prevent the `SIGPIPE` for just that send and return -1  (which is the expected behavior) and set errno to `EPIPE`.

This should be portable on POSIX systems, but `winsock2.h` doesn't support this flag, so some OS-specific code may be necessary.

- https://www.freebsd.org/cgi/man.cgi?query=send
- https://linux.die.net/man/2/send
- https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-send

**Caveats**

- My proposal is based on some research but I'm no expert on signal handling and this proposal should be evaluated by someone with more experience!
- Windows doesn't have signals, so this code would only apply to POSIX systems.

**Describe alternatives you've considered**

Globally ignoring the signal isn't appropriate for a library, since the programmer may need control of this signal. The same is true for a global signal handler, since it's not easy to reason about reentrancy.

Stumpless could delegate signal handling to the integrator (and provide documentation), but this feels like a detail that the library target should handle. Bottom line, a logging framework should never cause the process to be killed unexpectedly.

Another option:

Stumpless could treat the send to a network target as a critical section, block the signal while sending *on the current thread*, get pending signals, and restore the original signal mask.

1. Use `pthread_sigmask` to add `SIGPIPE` to the mask just prior to sending data. Store the original mask.
2. Send the data.
3. Use `sigtimedwait` to get pending `SIGPIPE` signals.
4. Restore the original mask.

This should work, but seems a lot more complicated than the solution I proposed. It could be a fallback approach if it turns out that I'm wrong about `MSG_NOSIGNAL`.

See the following for more detail: 

- https://linux.die.net/man/3/pthread_sigmask
- http://www.microhowto.info/howto/ignore_sigpipe_without_affecting_other_threads_in_a_process.html

**Additional context**

- Discussion: https://github.com/goatshriek/stumpless/issues/78#issuecomment-1229197064
- Runnable MRE: https://github.com/matthew-macgregor/stumpless-network-sigpipe-example

goatshriek:
Adds a `MSG_NOSIGNAL` flag to POSIX network send calls to avoid program termination via a `SIGPIPE` signal. While this was not an issue on Windows builds, Linux systems at a minimum were affected.

A roadmap item has been added for the next minor release to improve error handling so that a closed session is detected faster than a failed call to `send`, which may take longer than necessary in some cases.

This is intended to resolve #231 reported by @matthew-macgregor.

### DIFF
--- a/src/config/have_sys_socket.c
+++ b/src/config/have_sys_socket.c
@@ -243,7 +243,7 @@ sys_socket_sendto_target( struct network_target *target,
   result = send( target->handle,
                  msg,
                  msg_length,
-                 0 );
+                 MSG_NOSIGNAL );
   unlock_network_target( target );
 
   if( result == -1 ){
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -85,6 +85,7 @@ fixes, check out the
     * Object cache leak in some allocation failure scenarios
     * Creating network target with unsupported protocols
  - `close` syscall is no longer called on invalid socket handles
+ - TCP network targets do not hang when the session is closed by the receiver.
 
 ## [2.0.0] - 2020-12-27
 ### Added
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -10,6 +10,12 @@ or want to make a suggestion, please submit an issue on the project's
    as to a local file as well as a network server. Target chains will allow this
    stream to be defined as a logging target, and a logging call only made to
    this instead of manually logging to each target.
+ * [ADD] **Improved network target error detection**
+   Network targets do not currently detect errors that they would be able to in
+   some cases, such as with `select` or `poll`. This may lead to a connection
+   being left open for longer than necessary if the error is already detected.
+   This change will improve error detection in network targets to reduce the
+   time needed to pass these errors on to callers.
 
 
 ## 3.0.0 (next major release)
--- a/test/function/target/tcp4.cpp
+++ b/test/function/target/tcp4.cpp
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /*
- * Copyright 2019-2021 Joel E. Anderson
+ * Copyright 2019-2022 Joel E. Anderson
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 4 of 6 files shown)


src/config/have_sys_socket.c
```diff
@@ -243,7 +243,7 @@ sys_socket_sendto_target( struct network_target *target,
   result = send( target->handle,
                  msg,
                  msg_length,
-                 0 );
+                 MSG_NOSIGNAL );
```
matthew-macgregor, line 246: A quick note on this: it wasn't clear to me how this flag would behave on Windows (it's not listed as a flag in winsock2.h at least) and I didn't try it out while I was doing research on the issue.

goatshriek: Ah yes, thanks for pointing that out! I did run the test suites on a Windows machine before adding the fix, and winsock2 seems to return an error with send immediately (and doesn't end the program). This is different from POSIX systems with `sys/socket.h` where a send loop was needed to eventually capture the error (or cause termination before the flag was added). That is also why I added the roadmap to do some more error checking to get a faster "time to error" out of the POSIX side of things.

All that to say that the `src/config/have_winsock2.c` code that handles network target sending for Windows/Winsock builds is unchanged here for that reason. The test suite does still run on Windows builds just to make sure.

Issue #247, opened 2022-10-06
move wel config wrappers to separate header


goatshriek:
Stumpless uses a scheme involving `#define`names and macro functions to separate its portability measures from the application logic itself. This was originally done via a single header, `include/private/config/wrapper.h`, but as the library grew this single header became bloated. Now features are instead contained in their own dedicated wrapper headers. Older features need to be moved to their own as well to address the original problem.

Your task here is to factor out the Windows Event Log support wrappers into their own header. This is a good task for those with some C knowledge looking to learn about how to implement portability in a C application. As this is a refactoring task you can rely on the existing unit and integration tests as a safety net as you make changes and explore.

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, read through the [portability documentation](https://github.com/goatshriek/stumpless/blob/latest/docs/portability.md) to understand how portability is approached in the library.

Next, have a look at `include/private/config/wrapper/journald.h`. This header does the same wrapping for targets for systemd's journal service that the windows event log support should. This is a great file to copy as a starting point and modify to suit your needs. The new header should be `include/private/config/wrapper/wel.h`.

Finally, have a look at `tools/check_headers/stumpless.yml` and `tools/check_headers/stumpless_private.yml`. These files are used by a custom tool for header inclusion checks across the library's source code. For all symbols that you move to your new header file, you'll also need to change to point to your new header in these files. Any files that are present in `stumpless.yml` should be moved to `stumpless_private.yml` as well, as they were added before the public/private split in the yaml files was made. You'll see symbols for the journald wrappers already in `stumpless_private.yml` as an example.

Once you have moved things to your new header and updated the header checking tool manifests, you can use the header check tool to tell you what files need to have their `#include` statements updated. You can find examples of how to run the tool in the testing workflow that does it, `.github/workflows/analysis.yml`, or you can use the one-liner included below. You will need Ruby to run the tool.

```sh
tools/check_headers/check_headers.rb "include/**/*.h*" "src/**/*.c" "test/**/*.cpp" "docs/examples/**/*.c"
```

Good luck!

PR #263, opened 2022-10-14
Wrapper refactor wel

abhakta-47:
moving wel config wrappers to separate header #247


abhakta-47:
some wel related functions (eg. stumpless_close_wel_target, copy_wel_data etc) are present in `stumpless.yaml`, should they be exclusive to `stumpless_private.yaml` ?

include/private/target.h
```diff
@@ -96,4 +96,8 @@ COLD_FUNCTION
 int
 unsupported_target_is_open( const struct stumpless_target *target );
 
+static
+void
+close_unsupported_target( const struct stumpless_target *target );
+
```
goatshriek: This declaration is unnecessary as the static `close_unsupported_target` function is only needed/used within the `src/target.c` file. Please remove it.

goatshriek (changes requested):
Just one change to make to remove an unnecessary declaration, and then this should be all set!

The intended distinction between the `stumpless.yml` and `stumpless_private.yml` is for functions that are public facing (usually starting with `stumpless_`) and ones that aren't, like the `config_` functions. There are some private functions in the public manifest right now, but you don't have to worry about that change here - there will be a separate task later to finish fixing those up.

PR #267, opened 2022-10-16
Close #251: Set/Override procid for Entry

### LINKED ISSUE #251: set/override procid for entry
Stumpless currently derives the procid field of entries from the process ID it is executing in automatically. However, the field is a 128 printable ASCII field in RFC 5424, and users may want to set this id to something more meaningful to them. This task is to implement a pair of functions to set and retrieve the procid of an entry, which is used if it is set during entry formatting.

## General Approach
There are a few details left out of the following outline, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, read [this](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions) section of the development documentation on adding new functions. Once you understand how this process works, you'll be ready to define the new functions.

Next, browse over the implementations of the get and set functions for the `msgid` field of an entry. While these don't deal with a default behavior, their validation and error handling flow logic will be a useful starting point.

Add the declarations for your new functions into `include/stumpless/entry.h`. They should have the following signatures:

```c
STUMPLESS_PUBLIC_FUNCTION
const char *
stumpless_get_entry_procid( const struct stumpless_entry *entry );

// ...

STUMPLESS_PUBLIC_FUNCTION
struct stumpless_entry *
stumpless_set_entry_procid( struct stumpless_entry *entry, const char *procid );
```

The implementations of these functions should go into `src/entry.c`, and the tests for them in `test/function/entry.cpp`. Be sure that you take into account the following factors while you work on this:

 * the getter should return a newly allocated buffer holding the _effective_ procid, meaning if one has not been set then the pid should be used to make the string. Make sure this is mentioned in the documentation!
 * the setter should return the modified entry in the case of success, and NULL if not.
 * setting the procid to a NULL pointer should effectively restore default behavior. Note that this is different than msgid, which uses a NILVALUE in this case, as there is no default to fall back to.
 * the procid and supporting fields (such as a valid/not valid flag) should be stored in the entry struct itself, NOT as a pointer to dynamically allocated memory
 * validate supplied procids for their length and characters. `include/private/validate.h` has a number of functions you can use to do this.
 * documentation for the functions should include everything you see in other functions, including behavior, posix safety, and a `@since` tag for the next minor release.

Harry-Ramsey:
Overview
Implements the set/override functions for procid on entry structures.

Testing
TODO: Testing on Windows


goatshriek:
I haven't given this a full review, but in glancing through it I see you referencing the `id` field of the entry struct a few times. This field is _not_ the procid of the process, but is just a leftover field from an earlier design. It will likely be removed in the next major release, though that isn't on the roadmap yet. So, you don't want to modify/use this field.

Instead, I recommend you add a new character buffer to the entry structure as well as a boolean flag indicating if the procid has been overridden or not. Then use these two new fields to implement the new functionality.

Harry-Ramsey:
These changes should now be full implemented and hopefully it should be ready to be merged. I am unsure why the code coverage changes as everything has been commented.

include/private/validate.h
```diff
@@ -64,6 +64,16 @@ if( unlikely( ARG_NAME == NULL ) ) {                                           \
...
 
+/**
+ * Checks the char length of procid.
+ *
+ * @param the procid.
+ * 
+ * @return True if the procid is at or below the maximum allowed length, otherwise
+ * it will return false and raise the appropriate error.
+ */
+bool validate_procid_length(const char* procid );
```
goatshriek: add space before "const" here to be consistent with formatting

src/entry.c
```diff
@@ -976,6 +976,52 @@ vstumpless_set_entry_message( struct stumpless_entry *entry,
   return entry;
 }
 
+const char *
+stumpless_get_entry_procid( const struct stumpless_entry *entry ) {
+  char *procid;
+  
+  VALIDATE_ARG_NOT_NULL( entry );
+
+  procid = alloc_mem( sizeof(char) * 129 );
```
goatshriek: Use the `STUMPLESS_MAX_PROCID_LENGTH+1` expression here instead of a magic number. This is more readable, and most compilers will still convert this to a constant expression.

src/entry.c
```diff
@@ -976,6 +976,52 @@ vstumpless_set_entry_message( struct stumpless_entry *entry,
...
+}
+
+struct stumpless_entry *
+stumpless_set_entry_procid( struct stumpless_entry *entry, const char *procid ) {
+  VALIDATE_ARG_NOT_NULL( entry );
+
+  lock_entry( entry );
+
+  if( procid == NULL ) {
+    sprintf( entry->procid, "%d" , config_getpid() );
```
goatshriek: You can remove this line: don't bother modifying the procid field, just signal that it is not valid with the boolean.

src/entry.c
```diff
@@ -976,6 +976,52 @@ vstumpless_set_entry_message( struct stumpless_entry *entry,
...
+
+  if( procid == NULL ) {
+    raise_memory_allocation_failure();
+    return NULL;
+  }
+
+  lock_entry( entry );
+  
+  if( entry->procid_override == false ) {
+    sprintf( procid, "%d" , config_getpid() );
```
goatshriek: Avoid `sprintf` unless the formatting is non-trivial. In this case, you can use the code from `strbuilder_append_positive_int` or write a new helper in `include/private/inthelper.h`.

src/entry.c
```diff
@@ -976,6 +976,52 @@ vstumpless_set_entry_message( struct stumpless_entry *entry,
...
+
+  lock_entry( entry );
+
+  if( procid == NULL ) {
+    sprintf( entry->procid, "%d" , config_getpid() );
+    entry->procid_override = false;
+  }
+  else {
+    if( !validate_printable_ascii( procid ) || !validate_procid_length( procid ) ) return NULL;
+    sprintf( entry->procid, "%s" , procid );
```
goatshriek: `strncpy` will do here instead.

src/validate.c
```diff
@@ -39,6 +39,20 @@ static bool validate_string_length( const char* str, size_t max_length ) {
   return validation_status;
 }
 
+bool validate_procid_length(const char* procid ) {
+  size_t length = strlen( procid );
+  bool validation_status = true;
+
+  if( length > 128 ) {
```
goatshriek: use the `STUMPLESS_MAX_PROCID_LENGTH` value here as well instead of a raw integer.

goatshriek (changes requested):
There are a few changes listed above for you to make.

This is a good start, but I do not believe it will function as intended just yet. There have been no changes to the code that actually creates the message, and therefore the overriden procid, while reflected by the accessors, will not be used in actual messages.

Add a test that logs to a buffer target, and that checks the buffer afterwards for the string that you've set the procid to. This should fail. Write this test first, to make sure that it works correctly by failing. Then, update the implementation to correct the issue. I suggest modifying the code in `src/formatter.c` to check the overriden flag on the entry and append either the procid or the override buffer, depending on the value of the flag.

Harry-Ramsey:
I've tried to add the buffer tests but I have some issues when it comes to actually writing a message to the buffer. If you could just see what I missed, that would be great.

goatshriek:
the call to `stump_str` is redundant, the call to `stumpless_add_entry` sends the entry to the buffer. Try removing the call to `stump_str` and changing the substring check to the entry message.

Harry-Ramsey:
Thanks, hopefully the new test passes and can be merged into the branch now.

goatshriek:
This should be fine once the final remaining header check error is resolved.

Harry-Ramsey:
Hopefully all good to go! All changes have been made.

goatshriek (approved):

Issue #272, opened 2022-10-17
create wide string set entry message functions


goatshriek:
Windows applications that support unicode characters typically use wide UTF-16 strings to represent them, instead of the UTF-8 multibyte characters that stumpless supports by default. Wide string versions of some stumpless functions are available to allow users to avoid doing conversions themselves, but not all string-based functions support this. Your task is to add a wide string version of the entry set message function using a single string.

## General Approach
There are a few details left out of the following outline, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, read [this](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions) section of the development documentation on adding new functions. Once you understand how this process works, you'll be ready to define the new function.

Next, review the implementation of `stumpless_set_entry_message_str` in `src/entry.c`. You'll be using these as a template for your own implementation.

The last thing to review before you start your implementation is the wrapper defined in `include/private/config/wrapper/wstring.h` called `config_copy_wstring_to_cstring`, which can be used to get a copy of a wide string as a multibyte string. You'll use this to convert the supplied wide string to a multibyte string in your functions.

Once you've done this, you should be ready to implement the new function. It should have the following signature, declared in `include/stumpless/entry.h`

```c
STUMPLESS_PUBLIC_FUNCTION
struct stumpless_entry *
stumpless_set_entry_message_str_w( struct stumpless_entry *entry,
                                   const wchar_t *message );
```

Put the implementation of this functions in `src/entry.c`.

Finally, make sure that you test your new function! Tests should go in the `test/function/entry.cpp` suite, and should cover scenarios with NULL strings and invalid wide strings being passed in, in addition to normal expected usage.

maxmorenorebollar:
I'll try to solve this. It hasn't been fixed has it?

goatshriek:
Soumds good! It has not been resolved, though #281 was an attempt that went unfinished. I recommend reading through the changes and comments there as a starting point.

Nwogug:
Might take a crack at this as well

goatshriek:
[...]

PR #281, opened 2022-10-23
added a wide string version of the entry set message function using a single string

abhakta-47:
added a wide string version of the entry set message function using a single string as described in #272

### DIFF
--- a/include/stumpless/entry.h
+++ b/include/stumpless/entry.h
@@ -1461,6 +1461,39 @@ vstumpless_set_entry_message( struct stumpless_entry *entry,
                               const char *message,
                               va_list subs );
 
+/**
+ * Sets the message of a given entry.
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe. A mutex is used to coordinate changes to the
+ * entry while it is being modified.
+ *
+ * **Async Signal Safety: AS-Unsafe lock heap**
+ * This function is not safe to call from signal handlers due to the use of a
+ * non-reentrant lock to coordinate changes and the use of memory management
+ * functions to create the new message and free the old one.
+ *
+ * **Async Cancel Safety: AC-Unsafe lock heap**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, due to the use of a lock that could be left locked as well as
+ * memory management functions.
+ *
+ * @since release v2.1.0
+ *
+ * @param entry The entry to modify.
+ *
+ * @param message The new message to set on the entry. If this is NULL, then it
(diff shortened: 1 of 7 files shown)


abhakta-47:
`SetMessageWideStrTest.MallocFailureOnMessage` is failing.
I used existing `SetMessageStrTest.MallocFailureOnMessage` as template. Some tests are giving error code 24 (invalid encoding).
I am not sure what it is the issue maybe `set_malloc_result = stumpless_set_malloc( MALLOC_FAIL_ON_SIZE( 46 ) );` malloc size wouldn't be 46 ?

abhakta-47:
Also, how to create an invalid wide string to test for invalid wide string ?

goatshriek:
You may be seeing failures in the malloc test that you mentioned because of an issue with the function as a whole, not just the malloc failure portion. Be sure that you add a test that validates normal functionality, since right now all of your tests only cover error scenarios, and see if this fails as well. You may need to investigate the implementation of `config_copy_wstring_to_cstring` while troubleshooting, since this is the first significant usage of this function. There may be an error in that function's implementation.

For an invalid wide string, try constructing a string with invalid surrogates. You should be able to find examples of this in UTF-16 descriptions, including the [Wikipedia page](https://en.wikipedia.org/wiki/UTF-16).

abhakta-47:
```
#  ifdef HAVE_WINDOWS_H
#    include "private/config/have_windows.h"
#    define config_copy_wstring_to_cstring  windows_copy_wstring_to_cstring
#  elif HAVE_WCHAR_H
#    include "private/config/have_wchar.h"
#    define config_copy_wstring_to_cstring wchar_copy_wstring_to_cstring
#  else
#    include "private/config/fallback.h"
#    define config_copy_wstring_to_cstring  fallback_copy_wstring_to_cstring
#  endif
```

In `HAVE_WCHAR_H` case `wchar_copy_wstring_to_cstring` is working well.
So, I commented out that case to force use fallback funtions.
In fallback case there are linking errors during build.
`[build] /usr/bin/ld: libstumpless.so.2.1.0: undefined reference to 'fallback_copy_wstring_to_cstring'`

goatshriek:
That is due to the `src/config/fallback.c` module not being included during compilation as it is not needed by your system configuration. Since this wasn't used in your tests, it's unlikely to be the source of the problems you're seeing. Instead, try adding some print statements or debugging `wchar_copy_wstring_to_cstring` to see if it is behaving as you expect.

goatshriek:
Just checking in to see if this is still underway.

goatshriek:
Closing as stale.

PR #282, opened 2022-10-23
Factor config_getpid out of wrapper.h

acass91:
This PR is to move definition of config_getpid to its own header file. The details are discussed in issue https://github.com/goatshriek/stumpless/issues/271


goatshriek:
Looks good! Thanks very much for putting this together.

Issue #294, opened 2022-10-29
set/override hostname for entry


goatshriek:
Stumpless currently derives the hostname field of entries from the hostname of the machine it is running on automatically. However, the field is a 255 printable ASCII field in RFC 5424, and users may want to set this to something more meaningful to them. This task is to implement a pair of functions to set and retrieve the hostnameof an entry, which is used if it is set during entry formatting.

## General Approach
There are a few details left out of the following outline, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, read [this](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions) section of the development documentation on adding new functions. Once you understand how this process works, you'll be ready to define the new functions.

Next, review 4b6dc9a413de51375421c9902e52021ed0d58ceb which implemented this same functionality for the procid field of entries. This commit can serve as a template for the work you need to do here.

Add the declarations for your new functions into `include/stumpless/entry.h`. They should have the following signatures:

```c
STUMPLESS_PUBLIC_FUNCTION
const char *
stumpless_get_entry_hostname( const struct stumpless_entry *entry );

// ...

STUMPLESS_PUBLIC_FUNCTION
struct stumpless_entry *
stumpless_set_entry_hostname( struct stumpless_entry *entry, const char *hostname );
```

The implementations of these functions should go into `src/entry.c`, and the tests for them in `test/function/entry.cpp`. Be sure that you take into account the following factors while you work on this:

 * the getter should return a newly allocated buffer holding the _effective_ hostname, meaning if one has not been set then the machine hostname should be used to make the string. Make sure this is mentioned in the documentation!
 * the setter should return the modified entry in the case of success, and NULL if not.
 * setting the hostname to a NULL pointer should effectively restore default behavior.
 * the hostname and supporting fields (such as a valid/not valid flag) should be stored in the entry struct itself, NOT as a pointer to dynamically allocated memory
 * validate supplied hostnames for their length and characters. `include/private/validate.h` has a number of functions you can use to do this.
 * documentation for the functions should include everything you see in other functions, including behavior, posix safety, and a `@since` tag for the next minor release.

Gustrb:
I'm interested in tackling this issue. Is there any guid for beginners to contribute to the project? :)

Gustrb:
I'll look into implementing this functions! Thanks for such descriptive issue, it will be really helpful

PR #307, opened 2022-11-08
Set get for entry hostname

Gustrb:
This Pull Request is a response to the issue: #294
The hostname field is a 255 printable ASCII field in RFC 5424, and users may want to set this to something more meaningful to them. This task implements a pair of functions to set and retrieve the hostname of an entry, which is used if it is set during entry formatting.

### DIFF
--- a/include/stumpless/entry.h
+++ b/include/stumpless/entry.h
@@ -1552,7 +1552,8 @@ stumpless_set_entry_procid( struct stumpless_entry *entry, const char *procid );
  * appropriately.
  */
 STUMPLESS_PUBLIC_FUNCTION
-const char *stumpless_get_entry_hostname( const struct stumpless_entry *entry );
+const char *
+stumpless_get_entry_hostname( const struct stumpless_entry *entry );
 
 /**
  * Sets the hostname of a given entry. If hostname is NULL it will set hostname to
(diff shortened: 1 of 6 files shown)


include/stumpless/entry.h
```diff
@@ -1510,18 +1515,72 @@ stumpless_get_entry_procid( const struct stumpless_entry *entry );
...
+ * memory management functions.
+ *
+ * @param entry The entry to get the hostname of.
+ *
+ * @return The hostname of the entry if no error is encountered. If an error
+ * was encountered, then NULL is returned and an error code is set
+ * appropriately.
+ */
+STUMPLESS_PUBLIC_FUNCTION
+const char *stumpless_get_entry_hostname( const struct stumpless_entry *entry );
```
goatshriek: add a line break after the type, to match the formatting of other declarations

goatshriek (changes requested):
Just a few minor changes to make and this should be good to go. You'll need to add entries for your new functions to `src/windows/stumpless.def` to resolve the Windows build issues.

Also, check out the valgrind failure to see if there is a memory leak identified.

include/stumpless/entry.h
Gustrb: sure!

goatshriek:
Looks like there are still some memory leaks: you will need to free results provided by the get hostname function at a minimum, as this is freshly allocated each time. The valgrind output itself will give you feedback on where the problem is.

Gustrb:
how can I check the valgrind output locally??

goatshriek:
Assuming you have valgrind installed, you just need to run the test case with it once built. So, from the cmake build directory, something like `valgrind ./function-test-entry` should reproduce the error that the debug build shows.

Gustrb:
Hmmm, got it, thank you

goatshriek (approved):

goatshriek:
Thanks very much for putting this together, and responding to the changes quickly!

Gustrb:
I'm really glad to help!!

PR #310, opened 2022-11-21
seperate config_fopen.h wrapper

### LINKED ISSUE #298: factor config_fopen out of wrapper.h
Stumpless uses a scheme involving `#define`names and macro functions to separate its portability measures from the application logic itself. This was originally done via a single header, `include/private/config/wrapper.h`, but as the library grew this single header became bloated. Now features are instead contained in their own dedicated wrapper headers. Older features and functions need to be moved to their own as well to address the original problem.

Your task here is to factor the function used to open files into it's own header. This is a good task for those with some C knowledge looking to learn about how to implement portability in a C application. As this is a refactoring task you can rely on the existing unit and integration tests as a safety net as you make changes and explore.

## General Approach

97698be5b62c3fd1dddfd953db15a5c7265b896c is an implementation of a very similar task: you can use this commit as a template for this task.

First, read through the [portability documentation](https://github.com/goatshriek/stumpless/blob/latest/docs/portability.md) to understand how portability is approached in the library.

Next, have a look at `include/private/config/wrapper/wstring.h`. This header does the same wrapping for wide string support functions in the library. This is a great file to copy as a starting point and modify to suit your needs. The new header should be `include/private/config/wrapper/fopen.h`.

Finally, have a look at `tools/check_headers/stumpless_private.yml`. This file is used by a custom tool for header inclusion checks across the library's source code. You'll need to add an entry for `config_fopen` to this file, pointing to your newly created header.

Once you have moved things to your new header and updated the header checking tool manifests, you can use the header check tool to tell you what files need to have their `#include` statements updated. You can find examples of how to run the tool in the testing workflow that does it, `.github/workflows/analysis.yml`, or you can use the one-liner included below. You will need Ruby to run the tool.

```sh
tools/check_headers/check_headers.rb "include/**/*.h*" "include/**/**.h*" "src/**/*.c" "test/**/*.cpp" "docs/examples/**/*.c"
```

Good luck!

iltenahmet:
Continuing the elimination of include/private/config/wrapper.h by moving another function into its own wrapper header.
This fixes #298, which can be referenced for more details.


goatshriek (changes requested):
Thanks for putting this together! The core of this change looks like it should work out. There are a few changes for you to make before this can be tested and merged.

There is a merge conflict with the current default branch. You should be able to easily resolve this by merging in the latest branch and fixing up the includes manually.

Remove the `.DS_Store` files from the pull request. You can do this individually, or by adding a rule to the project `.gitignore`, this is up to you.

goatshriek (changes requested):
Thanks very much for updating with the latest branch! It looks like there are still a few changes needed.

You'll need to remove the DS_Store files that were added before the gitignore was updated. This should just be a matter of using `git rm --cached` to remove them. Also, consider updating the rule to `**/.DS_Store` so that it catches the files in any subdirectory.

The compilation issues are caused by `src/target/file.c` needing the new header added to it. You'll also want to update the includes based on the Static Analysis CI run where some unneeded headers can be removed after the refactoring.

goatshriek:
Just two more updates to go:

* the DS store files are still in there. You can check to see if you've removed them before pushing using the `git diff latest --name-only` command - once they're properly removed they won't show up.
* you can remove the `#include` statement for `private/config.h` from `include/private/config/wrapper.h`: after the refactoring, it is no longer needed.

goatshriek (approved):

goatshriek:
The changes look good!

For some reason the CI checks have not fired, but upon running them myself locally it looks like there is one more `#include` statement that can be removed after this refactoring:

`src/target/file.c: unused include private/config/wrapper.h`

My apologies for the back and forth on this. Once you make this final change, I believe this will be ready to merge in - making it myself and running the test suite locally showed no problems.

goatshriek:
Thanks very much for your persistence in getting this complete!

iltenahmet:
Thank you for being patient and helpful! This was my first pull request, I learned a lot.

PR #312, opened 2022-11-25
finish french translation

Msa360:
Completed the French translation todos.


goatshriek (approved):
Thanks very much for putting this together! The CI failures are due to some issues that are being resolved in #313, not caused by this change.

PR #326, opened 2023-01-09
add format checking of param values (#322)

### LINKED ISSUE #322: implement format checking of param values
The syslog RFC ([RFC 5424](https://tools.ietf.org/html/rfc5424)) specifies that param values only contain UTF-8 encoded messages. Checking the characters of param values will help users stay compliant with the standard, and also provide a point of mitigation for encoding bugs and vulnerabilities.

## General Approach

There are a number of functions that allow a param value to be supplied: the constructor for params, and accessor and mutator functions for entries and elements. Since this check will need to take place in several places, you will be best served by calling a single function wherever required.

`src/validate.c` contains the validation functions used within the library. You will need to define a new function that performs this validation in this module, and declare it in `include/private/validate.h`. Review the documentation on defining new functions [here](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions).

To understand what a valid UTF-8 string is exactly, you'll need to review [RFC 3629](https://www.rfc-editor.org/rfc/rfc3629). If that seems a bit overwhelming though, you can have a look at some code that already implements a UTF-8 validation check. This is in `test/helper/utf8.cpp` in the function `TestUTF8Compliance`. You can use this code as a reference, and be sure to raise an invalid encoding error when failures occur using the `raise_invalid_encoding` function. You'll also need to add a new error message: there is documentation about how to do this [here](https://github.com/goatshriek/stumpless/blob/latest/docs/localization.md#adding-new-strings).

After this change, add calls to your new function wherever a param value is passed as a parameter in the library, which are listed below.

 * `stumpless_new_param` (in `src/param.c`) This function is used by all others that create a new param, and so explicit checks are not needed in those.
 * `stumpless_set_param_value` (in `src/param.c`) This function is used by all others that modify a param value, and so explicit checks are not needed in those.

Update the documentation for the first two of these functions (creation and modification) to specify that the param value is restricted to UTF-8. The relevant documentation is in a doxygen comment format in the `include/stumpless/param.h` header file. Add the note about this to the `@param` descriptions of the param value parameters.

Don't forget to add tests for your new functionality! These should go alongside the tests for the functions that have param values. Find the existing tests for these functions as starting points, and add new tests that make sure that param values with valid UTF-8 are accepted, but non-UTF-8 strings are rejected.

momori256:
Resolves #322

Adds the validation of param values so that they follow the syslog RFC.
The RFC requires param values to be a valid UTF-8 string.

The UTF-8 validation function is `validate_utf8_compliance` in src/validate.c. It is called by
`stumpless_new_param` and `stumpless_set_param_value` in src/param.c.

### DIFF
--- a/include/private/config/locale/bg-bg.h
+++ b/include/private/config/locale/bg-bg.h
@@ -123,6 +123,10 @@
 " стойности между 0 и 7 включително"
 
 // todo translate
+#  define L10N_INVALID_STATE_DURING_UTF8_PARSING \
+"invalid state reached during UTF-8 string parsing"
+
+// todo translate
 #  define L10N_INVALID_TARGET_TYPE_ERROR_MESSAGE \
 "target type is incompatible with this operation"
 
--- a/include/private/config/locale/bn-in.h
+++ b/include/private/config/locale/bn-in.h
@@ -122,6 +122,10 @@
 "সংজ্ঞায়িত করা উচিত: 0 এর মধ্যে মান" \
 " এবং 7 সহ"
 
+// todo translate
+#  define L10N_INVALID_STATE_DURING_UTF8_PARSING \
+"invalid state reached during UTF-8 string parsing"
+
 #  define L10N_INVALID_TARGET_TYPE_ERROR_MESSAGE \
 "টার্গেট টাইপ এই অপারেশনের সাথে বেমানান"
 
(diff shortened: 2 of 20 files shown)


goatshriek:
I don't have time for a full review right now, but at a glance this seems like it should be good to go! I will give it a full review in the next 24 hours and either request changes or merge it in.

goatshriek (approved):
Thank you very much for putting this together, as well as updating the tests for coverage. Excellent work!

momori256:
Thank you for reviewing and merging!

PR #333, opened 2023-04-10
Update fr-fr.h

lilrawry:
correcting some mistakes & the to do section translated.

### DIFF
--- a/include/private/config/locale/fr-fr.h
+++ b/include/private/config/locale/fr-fr.h
@@ -114,9 +114,8 @@
 "les codes de sévérité doivent être définis conformément au RFC 5424:" \
 " valeurs entre 0 et 7 inclus"
 
-// todo translate
 #  define L10N_INVALID_STATE_DURING_UTF8_PARSING \
-"invalid state reached during UTF-8 string parsing"
+"état invalide atteint pendant UTF-8 string parsing"
 
 #  define L10N_INVALID_TARGET_TYPE_ERROR_MESSAGE \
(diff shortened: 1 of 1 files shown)


goatshriek:
Thanks for putting these updates in!

The static analysis is failing because it has detected a translation but still sees the `todo translate` comment on line 152. If you remove that comment from the string you've translated, it will be good to go.

Also consider translating the other string at line 119 (`L10N_INVALID_STATE_DURING_UTF8_PARSING`), which is the only other French string missing. If you decide to translate this as well, remember to remove the matching translate todo comment on line 117 as well.

lilrawry:
well received with thanks, i'll do that asap.

include/private/config/locale/fr-fr.h
```diff
@@ -151,7 +150,7 @@
 
 // todo translate
```
goatshriek: This todo line needs to be removed now that you've added a translation.

goatshriek (changes requested):
Almost there! Just one more todo comment to remove and it will be good to go.

include/private/config/locale/fr-fr.h
lilrawry: all good 👍🏼

goatshriek (approved):

lilrawry:
Hi again,
need anything else let me know 
keep me updated 
have a lovely day.

goatshriek:
All merged, and thanks again for the contribution!

lilrawry:
My pleasure 🤍✨

Issue #330, opened 2023-02-05
fuzz stump_str function


goatshriek:
In addition to a traditional testing suite, Stumpless also uses libFuzzer to perform fuzzing tests of some functions. Some functions do not yet have fuzzers implemented, despite being important functions. This task is to implement a fuzzer for the `stump_str` function.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

Note that in order to test your fuzzer, you will need to use `clang` as your compiler. If you have multiple compilers on your system, you can specify clang by adding arguments to cmake, such as `-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++`. Note that you'll also need to use `-DFUZZ=ON` to enable the fuzzing tests.

First, take a look at `test/fuzz/add_message_str.cpp`. You'll need to implement a function that is very similar to that in this file. Copy this file to `test/fuzz/stump_str.cpp`, and alter the contents to fuzz `stump_str`. Note that you'll still want to create the buffer target in your test, to avoid sending fuzzing tests into the default system log.

Next, in the `CMakeLists.txt` file, add an entry for your new fuzzing test using the `add_fuzz_test` function. Again, you can find where this is done for the `add_message_str` fuzzer and simply copy this and modify as needed.

Once this is done, make sure that you can run the new test. You can do this by building the `run-fuzz-test-stump_str` target, which is created by the `add_fuzz_test` function in the CMake build script you added earlier.

angiethompsonnn:
Hi, I plan on taking this up! The summary seems straightforward enough, but I will reach out if I have any questions.

angiethompsonnn:
@goatshriek I created a fork of the repo and have made a commit with the code changes I have made. I am a little unsure about how the buffer target is supposed to be used with stump_str(). I was able to run the fuzz test. I also am not sure how long the fuzzer is expected to run or if it runs indefinitely. 
[Fork commit: add stump_str fuzz test](https://github.com/angiethompsonnn/stumpless/commit/23614179d6f02df8b4588730b96f18d4f1ef5bfc)
![image](https://user-images.githubusercontent.com/114630075/231589722-d3604fdd-e712-42ba-a11b-d46d1e7a2ae6.png)

goatshriek:
[...]

PR #337, opened 2023-04-13
add stump_str fuzz test

angiethompsonnn:
Resolving issue: #330 
Added a fuzz test for the stump_str() function in test/fuzz. CMakeLists.txt has been modified to handle the test.


goatshriek:
Thanks very much for putting this together! Please consider reviewing #336 and providing any comments on improving the fuzz test documentation as well; it would be much appreciated.

PR #342, opened 2023-05-18
Add a npm installer package

ghost:
I tried downloading it through cmd and mingw using the debian and the C++ method, but it says connection failed for some reason. Please fix this as I have seen others do it


goatshriek:
I don't understand exactly what the request is, or how the title and description relate to the changes included in the pull request.

Can you share more details on what download methods you tried and how they failed? Screenshots or copied commands and their output would be most helpful.

I am unfamiliar with using NPM as a way to install a native library. Can you share an example of another project doing this, or more details on what you believe an "npm installer package" would do?

ghost:
Oh sorry. I did not know that. To get a NPM script you need node.js, which comes with NPM. To be honest this probably won't work since it works with JavaScript and this is C. But there is ruby in your code so there is a chance to make it using the gem install command. 

If you need more reference please check these two popular repositories that use NPM and gem.

- Rails https://github.com/rails/rails - An insanely big web-application framework that creates database-backed web applications. It uses Gem
- debug-js https://github.com/debug-js/debug - A javascript debugging utility. NPM

goatshriek:
Currently the installation packages provided are `.deb` for apt/dpkg based systems such as Debian and Ubuntu, `.rpm` packages for rpm-based systems such as CentOS and Fedora, a generic `.sh` installer for other Linux distributions, and a `.msi` installer for Windows systems. All of these are intended to provide quick and easy install experiences, so if you're having trouble with one of them please open a [bug report](https://github.com/goatshriek/stumpless/issues/new?template=bug_report.md) with details of the failure and I'll address it.

> I tried downloading it through cmd and mingw using the debian and the C++ method, but it says connection failed for some reason.

Are you able to share what these steps were and what the failure was? I'd like to fix any problems that exist with the `.deb` packages.

If you have a request for another package manager, please open a [feature request](https://github.com/goatshriek/stumpless/issues/new?template=feature_request.md) describing what you want and we will see if we can add it. Managers for other languages such as npm and RubyGems are not something we'll be able to support, but there are many others that should be possible.

> But there is ruby in your code so there is a chance to make it using the gem install command.

The Ruby code in the project is used for development and build steps, and isn't part of the library itself. Making a gem for this project wouldn't be a natural install option, since it isn't a Ruby-based tool. A Ruby runtime isn't even a requirement for using the library itself, only during development, and for the build of the C++ bindings.

ghost:
In that case, is there a way to download it using the command line? Or is there any way I can fix this? (Sorry for the long wait, I had some work to do.)  I am a web developer, so I won't know as much. But I did some research and found this- 

image
Maybe there is something similar?  If there is please list it.

goatshriek:
Right now the best command-line installation method is to build from the source, following along with the steps outlined in the [install guide](https://github.com/goatshriek/stumpless/blob/latest/INSTALL.md). If you notice missing info in this please let me know: it should be good for you as long as you have cmake and any build toolchain installed.

If there is a specific package manager you would like to use in the future, please open a separate issue for it and I'll see about getting the project in there.

ghost:
I saw the installation guide and I didn't see anything from the source. But it seems as though there is a `msi` format that I can download. So since this has UI, maybe this won't work on the command-line. 

>  If you notice missing info in this please let me know: it should be good for you as long as you have cmake and any build toolchain installed.

I am not familiar in C or CMake. I do have knowledge to install JavaScript on the command-line. Can you provide some information on CMake? I do have a build toolchain- Gulp.

Anyway, my problem is to check the debian and and the C++ method. Otherwise, please give me the installation script if you want me to point out the problem.

goatshriek:
Since Gulp is a Javascript build tool, it's unlikely to help here. CMake serves a similar purpose, typically for C and C++ projects. CMake generates build scripts for various C compilation toolchains such as Make and GCC, Visual Studio, and others. If you're working in MinGW you probably already have these installed, so the build from source commands should work.

If you need a command-line installation method for the MSI, you can use [msiexec](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/msiexec) to accomplish this. Windows PowerShell will let you do an `Invoke-WebRequest`, similar to wget, to download the file as well. So for Windows, you could do these two commands in a PowerShell prompt with appropriate permissions:

```ps1
Invoke-WebRequest -Uri https://github.com/goatshriek/stumpless/releases/download/v2.1.0/stumpless-cpp-2.1.0-win64.msi -OutFile stumpless-cpp-2.1.0-win64.msi
msiexec /i stumpless-cpp-2.1.0-win64.msi
```

If you need a set of commands to install on Debian, you could do this with a wget followed by dpkg to accomplish the same thing:

```sh
wget https://github.com/goatshriek/stumpless/releases/download/v2.1.0/stumpless-cpp-2.1.0-amd64.deb
sudo dpkg -i stumpless-cpp-2.1.0-amd64.deb
```

ghost:
I tried the PowerShell code. Yes, it did install the `msi` package and it brought the setup wizard and it did start installing. The problem here is that after I gave it administrator privileges, it crashed. I tried again and it also crashed. And the debian way again said "connection failed" like before. 

Like you said, I installed CMake. But can the command work in PowerShell? I am not as familiar with MinGW than other 
 command-lines. I don't know how CMake works since this is my first time using it.

goatshriek:
Can you share a screenshot of either failure?

CMake itself can run from PowerShell if you have installed it to your system path, but it may not find your compiler. Seeing the commands you've tried and their output is probably the best next step.

ghost:
Sure. Here is a screenshot of the failure-
image
By the way, I've seen this happen many times before. For example, I was trying to install a calculator but the same thing happened. Either way, this may be happening because I have Windows 11 and this is not compatible with it.

I do not have much experience with PowerShell. I have read through how to use CMake and I'm confident that I can use it without PowerShell.

goatshriek:
Hmm, I can reproduce that error by running the command with a file that doesn't exist. Can you share the output of `dir` in the same folder that you executed `msiexec` from? I'm assuming this is the same folder as the `msi`. It should look something like this:

```bat
stumpless> dir
 Volume in drive C has no label.
 Volume Serial Number is <...>

 Directory of C:\Users\<username>\Downloads

05/25/2023  07:23 AM    <DIR>          .
05/25/2023  07:23 AM    <DIR>          ..
05/23/2023  05:41 AM           532,480 stumpless-cpp-2.1.0-win64.msi
               1 File(s)        532,480 bytes
               2 Dir(s)  <...> bytes free

stumpless> msiexec /i stumpless-cpp-2.1.0-win64.msi

stumpless>
```

ghost:
Yes, it does look kind of like that. Unfortunately, the code has been looking a lot different. It even looks like it's supposed to be correct and the normal information that should be given. But something popped up that looks like can be related to the code (I am assuming this because the installer wants to be administrator and this is something related to it.

  
image
Powershell was already administrator before so this must be a copy created to it.

Anyway, I have another theory on why this is not working. It maybe is a 32-bit application. Windows 11 does not support 32-bit applications. I think your code got mixed up.

goatshriek:
There is both a 32-bit and 64-bit version of the installer - if you've used the [64-bit MSI](https://github.com/goatshriek/stumpless/releases/download/v2.1.0/stumpless-cpp-2.1.0-win64.msi) then there should be no architecture incompatibility.

goatshriek:
Closing due to deleted requestor account.

PR #344, opened 2023-06-01
danish localization

LasseEmilBuch:
added the danish langauge localization

my first real contribution to an opensource project so hopefully it works :)


goatshriek:
I can provide a full review in a few days when I have an environment I can work in. In the meantime, please verify that the header file is in UTF-8 encoding (it may be, I just can't check), as that may be the reason for the failing l10n check in the static analysis step.

CMakeLists.txt
```diff
@@ -203,6 +203,9 @@ elseif(LOCALE MATCHES "^hi[_-]")
 elseif(LOCALE MATCHES "^bn[_-]")
   set(STUMPLESS_LANGUAGE "bn-IN")
   set(USE_LOCALE_BN_IN TRUE)
+  elseif(LOCALE MATCHES "^da[_-]")
+  set(STUMPLESS_LANGUAGE "da-DK")
+  set(USE_LOCALE_BN_IN TRUE)
```
goatshriek: this should be `USE_LOCALE_DA_DK` instead

.github/workflows/locale.yml
```diff
@@ -290,3 +290,24 @@ jobs:
         run: |
           make check
           if grep "DEPRECATED" Testing/Temporary/LastTest.log; then exit 1; fi
+  linux-da-dk:
+    name: "linux, da-dk"
+    runs-on: "ubuntu-20.04"
+    env:
+      LANG: "da_dk.UTF-8"
```
goatshriek: I don't think the casing will cause anything to break, but best to change this to `da_DK.UTF-8` for consistency with the other tests.

include/private/config/locale/da-dk.h
```diff
@@ -0,0 +1,257 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** @file
+ * United States English localization strings.
+ */
+
+#ifndef __STUMPLESS_PRIVATE_CONFIG_LOCALE_EN_US_H
+#  define __STUMPLESS_PRIVATE_CONFIG_LOCALE_EN_US_H
```
goatshriek: the header guard should change to `__STUMPLESS_PRIVATE_CONFIG_LOCALE_DA_DK_H`

include/private/config/locale/da-dk.h
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** @file
+ * United States English localization strings.
```
goatshriek: Update this to describe your danish translation.

README.md
```diff
@@ -28,7 +28,7 @@ easier:
  * completely thread safe
  * can be adjusted or removed during compilation for zero runtime impact
  * localized for multiple languages :brazil: :bulgaria: :cn: :czech_republic:
-   :de: :es: :fr: :greece: :india: :it: :poland: :slovakia: :sweden: :us:
+   :de: :es: :fr: :greece: :india: :it: :poland: :slovakia: :sweden: :us: :dk:
```
goatshriek: the new flag should be `:denmark:` according to github markdown flavor

goatshriek (changes requested):
Thank you very much for putting this together! If you're able to resolve the comments in this review, as well as the two additional points, I believe this will be ready to merge. You can disregard the CI failures regarding codecov uploads: those aren't your fault.

The new `da-dk.h` file appears to be encoded in ISO-8859 instead of UTF-8, which can cause issues. Please convert it to UTF-8. I was able to do this on a linux machine using the command `iconv -f ISO-8859-15 -t UTF-8 da-dk.h > utf8-da-dk.h`, but of course you should probably check the result of that to make sure it's still correct.

Once the encoding is correct the l10n check script will run to completion, but for me it identified a few lines that are too long. I've included the output for you below so that you can go ahead and correct them: you'll just need to split them up to be a little shorter.

```
utf8-da-dk.h: line "en MULTI_SZ registreringsdatabase værdi var hverken tom eller Afsluttet med to NULL's" \ is longer than 80 characters
utf8-da-dk.h: line "alvorligheds koder skal være defineret i Overensstemmelse med RFC 5424: værdier mellem 0" \ is longer than 80 characters
```

include/private/config/locale/da-dk.h
LasseEmilBuch: unsure what you mean by this, rest should be done though

LasseEmilBuch:
going on holiday for the weekend so if this new version doesnt work youre either gonna have to wait until monday or solve the remaining issues yourself

sorry for the rookie mistakes, this is my first real time contributing to something like this :)

include/private/config/locale/da-dk.h
goatshriek: Instead of the `@file` directive saying "United States English localization strings", it should say something like "Danish localization strings".

.github/workflows/locale.yml
```diff
@@ -290,3 +290,24 @@ jobs:
...
+    env:
+      LANG: "da_DK.UTF-8"
+    steps:
+      - uses: actions/checkout@v3
+      - name: Configure
+        run: |
+          cmake .
+      - name: Check Language
+        run: |
+          grep -q "STUMPLESS_LANGUAGE \"da_DK\"" include/stumpless/config.h
```
goatshriek: Sorry I missed this on the first time through: this grep should look for "da-DK" (note the '-' instead of the '_'), which should resolve the CI failure.

goatshriek (changes requested):
No worries at all! This is not an emergency, so please don't feel rushed. It's fine to take a few days to get back around to it.

Just a few more noted tweaks, and it looks like the ISO vs. UTF encoding issue remains. You should be able to do a `file include/private/config/locale/da-dk.h` to see if it is reported as UTF-8, as a sanity check. You an also directly execute the localization check to make sure the issue is resolved before pushing via `scripts/check_l10n.rb "include/private/config/locale/*-??.h"`, though of course you will need Ruby installed for that to work.

goatshriek:
Thanks again for putting this together, especially sticking with it through the encoding issues!

Issue #273, opened 2022-10-17
test wel insertion string with invalid utf-8


goatshriek:
The test suite for the Windows Event Log is missing tests that cover some scenarios, partially due to a lack of easy access to coverage information for Windows-enabled builds. This task is to add a test covering behavior of the library in some scenarios.

## General Approach
There are a few details left out of the following outline, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

The function you will be testing is `stumpless_set_wel_insertion_string`, which allows the specification of a string to use in place of an insertion string placeholder. The function accepts a UTF-8 (multibyte) string, which will later be converted to a UTF-16 (wide) string during usage.

Look at the existing tests for this function in `test/function/config/wel_supported.cpp`. You'll add new tests for this function in this file, where the insertion string provided is an invalid UTF-8 string. Be sure that you also test that the insertion string in the provided index is not modified when a bad string is provided, in scenarios including it is already assigned, it is out of range, and it is in range but not yet assigned.

Boombag0607:
@goatshriek  I would like to work on this, if it is still up for the taking. Also can you provide other UTF-8 string references just for clarification?

goatshriek:
Sure thing! The [wikipedia entry](https://en.wikipedia.org/wiki/UTF-8) is a solid place to look for a start. One of the easiest invalid states to make is an unexpected continuation byte.

PR #346, opened 2023-06-14
Add tests for handling invalid UTF-8 insertion strings in WelSupportedTest

Boombag0607:
## Description

This pull request addresses issue #273. Adds new tests to the `WelSupportedTest` fixture in order to handle invalid UTF-8 insertion strings. The tests focus on various scenarios to ensure correct behaviour and error handling.

### Test Cases

1. `InvalidUtf8InsertionString`: This test verifies that an error is raised when an invalid UTF-8 string is provided as an insertion string.

2. `InsertionIndexNotModifiedWhenInvalidUtf8StringProvided`: This test checks that the insertion string at the specified index remains unchanged when an invalid UTF-8 string is provided.

3. `InvalidUtf8InsertionStringOutOfRange`: This test ensures that an error is raised when attempting to set an invalid UTF-8 string at an out-of-range index.

4. `InsertionIndexNotModifiedWhenOutOfRangeIndexProvided`: This test verifies that the insertion string at a valid index remains unchanged when an out-of-range index is provided for an invalid UTF-8 string.

5. `InsertionIndexNotModifiedWhenInRangeButNotAssigned`: This test confirms that an unassigned insertion string at a valid index returns `nullptr`.

### DIFF
--- a/.github/workflows/cygwin.yml
+++ b/.github/workflows/cygwin.yml
@@ -1,10 +1,7 @@
 name: "Cygwin"
 
-on:
-  push:
-    branches: [ latest ]
-  pull_request:
-    branches: [ latest ]
+on: 
+  workflow_dispatch
 
 env:
   CTEST_OUTPUT_ON_FAILURE: 1
--- a/.github/workflows/cygwin.yml
+++ b/.github/workflows/cygwin.yml
@@ -1,7 +1,10 @@
 name: "Cygwin"
 
-on: 
-  workflow_dispatch
+on:
+  push:
+    branches: [ latest ]
+  pull_request:
+    branches: [ latest ]
 
 env:
   CTEST_OUTPUT_ON_FAILURE: 1
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,7 +1,10 @@
 name: "Windows"
 
-on: 
-  workflow_dispatch
+on:
+  push:
+  pull_request:
+    branches:
+      - latest
 
 env:
   CTEST_OUTPUT_ON_FAILURE: 1
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,10 +1,7 @@
 name: "Windows"
 
-on:
-  push:
-  pull_request:
-    branches:
-      - latest
+on: 
+  workflow_dispatch
 
 env:
   CTEST_OUTPUT_ON_FAILURE: 1
--- a/test/function/config/wel_supported.cpp
+++ b/test/function/config/wel_supported.cpp
@@ -336,23 +336,31 @@ namespace {
   }
 
   TEST_F( WelSupportedTest, InvalidUtf8InsertionString ) {
+    const struct stumpless_error *error;
     const struct stumpless_entry *entry_result;
     LPCSTR invalid_utf8_string = "\xc3\x28 invalid";
 
-    entry_result = stumpless_set_wel_insertion_string( insertion_entry, 0, invalid_utf8_string );
+    entry_result = stumpless_set_wel_insertion_string( insertion_entry, 
+                                                        0, 
+                                                        invalid_utf8_string );
+    EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
     EXPECT_NULL( entry_result );
   }
 
   TEST_F( WelSupportedTest, InsertionStringNotModifiedWhenInvalidUtf8StringProvided ) {
+    const struct stumpless_error *error;
+    const struct stumpless_entry *entry_result;
     LPCSTR valid_string = "valid string";
     LPCSTR invalid_utf8_string = "\xc3\x28 invalid";
-    const struct stumpless_entry *result;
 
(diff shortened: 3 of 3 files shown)


goatshriek:
I'm converting this to a draft for now just to make it clear that this is still in work. Once you're all finished with it or you need help, let me know and I'll review it.

Boombag0607:
Sure thing. The problem I'm facing is that I'm currently using a Mac M1 and since the test is particularly Windows related, I can't really find a way to compile the files/ run the workflows locally. There's really no way for me to know the checks before pushing. Can you suggest a workaround?

goatshriek:
Without having access to a Windows machine to work on, this will be a little more difficult for you. One possible workaround is to enable Github Actions on your fork, remove all workflows but the Windows ones, and push to it to run the Windows CI workflows when you need to test something. You'd need to be sure to add the original workflow files before the request is marked as ready.

You could also try something like [Wine](https://www.winehq.org/), but the library hasn't been tested on this and I wouldn't be able to offer any support for getting it working. Using Actions on your own fork is probably the easiest thing to do, so that you won't need to wait for me to approve runs on the upstream repository.

.github/workflows/cygwin.yml
```diff
@@ -1,10 +1,7 @@
 name: "Cygwin"
 
-on:
-  push:
-    branches: [ latest ]
-  pull_request:
-    branches: [ latest ]
+on: 
+  workflow_dispatch
```
goatshriek: This file shouldn't be changed - restore it to the original version now that your testing is complete.

.github/workflows/windows.yml
```diff
@@ -1,10 +1,7 @@
 name: "Windows"
 
-on:
-  push:
-  pull_request:
-    branches:
-      - latest
+on: 
+  workflow_dispatch
```
goatshriek: This file should also be restored to the original so that it isn't included in this change.

test/function/config/wel_supported.cpp
```diff
@@ -335,6 +335,47 @@ namespace {
     EXPECT_ERROR_ID_EQ( STUMPLESS_ARGUMENT_EMPTY );
   }
 
+  TEST_F( WelSupportedTest, InvalidUtf8InsertionString ) {
+    const struct stumpless_entry *entry_result;
+    LPCSTR invalid_utf8_string = "\xc3\x28 invalid";
+
+    entry_result = stumpless_set_wel_insertion_string( insertion_entry, 0, invalid_utf8_string );
```
goatshriek: This line and a few others are over 80 characters long: reformat where you can to bring these to 80 characters or under. There are examples elsewhere in this file of how to format long function calls across multiple lines.

test/function/config/wel_supported.cpp
```diff
@@ -335,6 +335,47 @@ namespace {
...
+    EXPECT_NULL( entry_result );
```
goatshriek, line 347: Add checks after calls with invalid UTF8 to make sure that the error id is set correctly: it should be `STUMPLESS_INVALID_ENCODING`. See some of the above tests for examples of how to do this using the `EXPECT_ERROR_ID_EQ` macro. Also note that this helper requires a `stumpless_error` struct named "error" to be declared in order to work.

test/function/config/wel_supported.cpp
```diff
@@ -335,6 +335,47 @@ namespace {
...
+  /* 
+    Add the test case for invalid utf8 insertion string when insertion string is out of range
+  */
+
+  TEST_F ( WelSupportedTest, InsertionIndexNotModifiedWhenInRangeButNotAssigned ) {
+    const struct stumpless_entry *result;
+    LPCSTR invalid_utf8_string = "\xc3\x28 invalid";
+    LPCSTR valid_string = "valid string";
+
+    stumpless_set_wel_insertion_strings(insertion_entry, 2, invalid_utf8_string, NULL);
```
goatshriek: add result and error id checks after this call

goatshriek (changes requested):
Thanks very much for putting these tests together, they are mostly ready! There are a few updates for you to make listed below, after which I will merge this in.

.github/workflows/cygwin.yml
Boombag0607: Oh that was not intended to happen, must have been merged from a different branch.

test/function/config/wel_supported.cpp
Boombag0607: Sure, I'm fixing the same. Can you elaborate on the insertion string being out of range part (from the issue #273) so that I can add this test to the commit as well, I could not get how a valid insertion string be out of range and the invalid one being in range.

goatshriek: The language in the issue is a little confusing, sorry about that. You can achieve the state I was trying to describe by setting a valid insertion string at a high index (I would use index 3 or 4, since the fixture entry already has 2 insertion strings), and then setting the invalid one at index 2. Index 2 won't have been assigned yet, but will be in the already allocated range because of the higher string successfully set. Your test will need to make sure that it is still empty after the attempt to set it to an invalid string.

Boombag0607: Got it.

goatshriek (approved):

PR #351, opened 2023-07-06
adding Hebrew localization

AssafBardugo:
For the issue: implement hebrew localization #350


goatshriek (approved):
This change looks good! The failed check is due to a localization script regex that doesn't support string literal concatenation. I have a tentative fix in the `fix-l10n-check` branch, which should resolve it. Once that's merged in, I'll merge in this pull request.

goatshriek:
Thanks again for putting this together and your patience in waiting for the merge!

Issue #193, opened 2021-10-10
refactor tests for invalid format items


goatshriek:
Tests for names and other elements with formatting errors use much of the same code, and could be made more compact by refactoring the common code into helper test functions. Some documentation on creating and using test helpers can be found in [test/README.md](https://github.com/goatshriek/stumpless/blob/latest/test/README.md) and [docs/test.md](https://github.com/goatshriek/stumpless/blob/latest/docs/test.md)

See the discussion of #184 for the context behind this request.

Marecek460:
Hello. I would like to help work on this issue.

goatshriek:
Absolutely! I have assigned the issue to you to mark it as in work. This issue is admittedly a little vague, so please reach out if you have any questions or need clarification.

goatshriek:
Hello @Marecek460, just checking to see if you still intend to work on this? No worries if not, just want to see if this should be released for others.

goatshriek:
This item is open for others to tackle once more.

goatshriek:
Resolved by #474.

PR #356, opened 2023-08-09
(#193) Refactor tests to reduce common code

ade3p:
Hi @goatshriek, based on my understanding of the discussion in https://github.com/goatshriek/stumpless/pull/184, https://github.com/goatshriek/stumpless/issues/193 - have added a helper macro basically to hold the invalid name strings which are then used as source instead of adding it individually. So have updated for the three files in that closed PR, please check and let me know if it adds some value.


goatshriek (changes requested):
Thank you very much for putting this change together! It is a welcome addition.

Your approach to collect the invalid strings into a single place is a good one. However, I would like you to change the setup so that invalid param names are not stored in the source code, but are instead stored in separate files in the `test/corpora` folder in a new directory named `invalid_param_name`. You can then write a helper that iterates through the files in this directory and uses the `load_corpus` function that already exists in `test/helper/fixture.hpp` to get each test string. This header is probably the better place to put your new helper (and the corresponding fixture.cpp source for implementation), instead of creating a new header.

The reasoning here is so that future fuzzing tests that need parameter names can be pointed at this corpus for a faster ramp up. It will also allow discovered crashes to be added to this folder to immediately improve both fuzzing efforts and unit tests.

It is a relatively new structure, so I wouldn't have expected you to reach this conclusion, and I apologize for you not having a way to anticipate this. I plan to improve the documentation around this part of the codebase when I update the fuzzing tests to use Google's [fuzztest](https://github.com/google/fuzztest), which is a currently maintained successor to the libFuzzer setup used today.

ade3p:
👋 Hi! based on the comment above, made following tweaks:

- [x] moved the source strings to `test/corpora/invalid_param_name` as different files
- [x] added helper method `load_corpus_folder` under `test/helper/fixture.hpp` which basically iterates the mentioned directory and reads the contents using `load_corpus` function. Returning `stumpless_test_data` pointing to the array and the corresponding size
- [x] from the tests, the helper is used to get the source strings 
- [x] removed `test_strings.hpp` added earlier

goatshriek (changes requested):
First of all, thanks very much for continuing to work on this! It looks like it is nearing completion.

It looks like there are a few issues with the tests failing, which I haven't looked at in too much detail. I'll re-run the tests once you work out a fix.

You'll also need to update the header manifests in `tools/check_headers/stumpless.yml` with `load_corpus_folder` and `tools/check_headers/standard_library.yml` with `opendir`, as well as add a missing header to `test/function/fixture.cpp`. That will fix the static analysis header check that is failing.

goatshriek:
Just checking in to see if you still continue to work on this? No worries if not.

goatshriek:
Closing [due to inactivity](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md#you-still-there).

Issue #367, opened 2023-09-16
update stumpless_element_to_string format


goatshriek:
The `stumpless_element_to_string` function was originally implemented with a relatively arbitrary string format similar to: `<name>:[...]` . However, as the [CLI logger](https://github.com/goatshriek/stumpless-logger) is developed, and the `stumpless_new_element_from_string` function will soone be implemented, there is a strong case for this format to instead be `name=[...]`. This change adjusts the function, it's documentation, and tests to this new format.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

There are three main things that need to change for this adjustment: the implementation itself, the documentation, and the test suite. The [development documentation](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) has everything that you need to get started and more. This change is relatively simple, so you won't need to do much more than build the library and run the test suite.

Starting with the implementation will be most straightforward approach. The function is implemented in `src/element.c`. The adjustments to match the new format should be relatively minor.

After changing the implementation, the tests will be failing. You can find them in `test/function/element.cpp`. Run the test suite (using the `check` target is the easiest way to do this). Fix anything that has broken after your update.

Finally, update the documentation in `include/stumpless/element.h` for `stumpless_element_to_string`.

You may find #366 interesting to review, as it implemented a very similar change for the param to string format.

rmknan:
Can you assign this issue to me? I had finished it earlier. I can commit this along side the #293 and maybe you can merge both together?

goatshriek:
That sounds fine! Please open a new pull request for this change though. This work is unrelated to that in the PR you currently have opened, and having it end up in a separate commit will ease future tasks that might arise like merge conflicts or reverting changes.

PR #369, opened 2023-09-16
#367 update stumpless_element_to_string format

rmknan:
Updated the format of `stumpless_element_to_string` in the file` src/element.c`, the test suite and the documentation.


rmknan:
#367

rmknan:
I don't understand why it still shows 12 commits.

rmknan:
When i `git pull upstream latest`. It says it is updated.

goatshriek:
The default target changes have just been merged - redoing the pull from the upstream latest should now remove those changes from this one.

When you create a new branch for a PR, make sure that you are basing it off of the upstream's target branch (`latest` for this project). That will ensure that you don't have local commits from other changes showing up in your pulls. This can be especially problematic when commits are squashed when merged, as they have been for your submissions.

src/element.c
```diff
@@ -246,16 +246,13 @@ stumpless_element_to_string( const struct stumpless_element *element ) {
...
-      format[name_len + 3] = '[';
+      // name=[param_1_to_string,param_2_to_string,etc.] (with params)
+      format[name_len ] = '=';
+      format[name_len + 1] = '[';
       format[pos_offset] = ']';
     } else {
-      // <name> (no params)
+      // name (no params)
       // pos_offset is name_len + 4 here
       pos_offset -= 3;
```
goatshriek: This offset needs to be adjusted to account for the change in character position.

goatshriek (changes requested):
There are still some offsets that are incorrect as a result of the changes. Please also update the computation of `format_len` on lines 221 and 224 as well - they need to be updated based on the number of characters changing.

src/element.c
rmknan: Is it correct now?

goatshriek: It seems like there are still some tests failing, which may be because some of these counts are still off.

rmknan: Is there any way to verify the changes locally? The build test runs fine, so I am unable to tweak it.

goatshriek: I recommend you look through the failed CI checks, particularly for the MacOS builds, to see what tests failed there. Different compilers have different sanitization strategies, so what works in GCC may not work in clang, for example. You can also try valgrind, which is also failing in the Linux CI runs, which will point you at more specific memory access violations.

rmknan: Can I run these CI checks locally? or would I have to commit and then check here?

goatshriek: If you want to run the checks locally (which I definitely recommend), you can use `make check` to build all of the tests first. Then, use `valgrind` to run the specific test that is failing in the CI. You can see how the CI tests run this in the `.github/workflows/linux.yml` file around [line 56](https://github.com/goatshriek/stumpless/blob/latest/.github/workflows/linux.yml#L56) as an example. Once you make changes to the test/source, don't forget to re-run the make command to rebuild the tests with your changes.

rmknan: I get this when I try to run Valgrind, any ideas on whats causing this?

```
==11209== Memcheck, a memory error detector
==11209== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==11209== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==11209== Command: ./function-test-buffer
==11209== 
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x23
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x23
==11209== Valgrind: debuginfo reader: ensure_valid failed:
==11209== Valgrind:   during call to ML_(img_get)
==11209== Valgrind:   request for range [1956341782, +4) exceeds
==11209== Valgrind:   valid image size of 2202248 for image:
==11209== Valgrind:   "/home/rmknan/Github/Stump/build/function-test-buffer"
==11209== 
==11209== Valgrind: debuginfo reader: Possibly corrupted debuginfo file.
==11209== Valgrind: I can't recover.  Giving up.  Sorry.
==11209== 
```

goatshriek: I am not sure what would cause that offhand. Perhaps a difference in valgrind debug info support and the compiler used?

rmknan: How would I check if these two are the same or not? and please be as detailed as possible. Thank you.

goatshriek: I'm not sure how you would check compatibility between valgrind and the dwarf info. A search for part of the error messages turns up [this thread](https://github.com/llvm/llvm-project/issues/56550), which might be worth trying. You can set the flags during compilation by setting the `CFLAGS` environment variable before running cmake, if you need to.

goatshriek (approved):
Thanks very much for sticking with this despite the many bumps! It looks like everything is good to go now. I'll merge this in the next day or so.

Issue #360, opened 2023-08-25
param from string


goatshriek:
The Stumpless [Rust crate and CLI tool](https://github.com/goatshriek/stumpless-logger) creates targets and entries to do whatever logging needed. In order to provide CLI support, many of these tasks need to use the strings provided in the command to set up stumpless structures. One of these is the string used to create structured data params of the logged entry. To support this, a new function is needed named `stumpless_new_param_from_string` which parses a string and attempts to create a param from it.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First read the relevant section of [RFC 5424](https://www.rfc-editor.org/rfc/rfc5424#section-6.3.3) and the [logger manpage](https://man7.org/linux/man-pages/man1/logger.1.html) section for the `--sd-param` option. The RFC describes what characters are valid in param names and values. These are already validated by the `stumpless_new_param` function, so you will be best served by simply using this function to create your new param to take advantage of the checks. The manpage describes the string that is used to create params. You will expect strings in this same format: `name="value"`.

Optionally, you may review the current [Rust implementation](https://github.com/goatshriek/stumpless-logger/blob/latest/src/main.rs#L340) of this functionality for the CLI tool. This change will move this functionality into the C library so that other language bindings such as C++ can offer the same functionality.

Next, read [this](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions) section of the development documentation on adding new functions. Once you understand how this process works, start defining the new function.

Your new function will be named `stumpless_new_param_from_string`, should take a single string parameter, and should return a new `stumpless_param` structure based on the string. In the event of an error (such as a NULL or invalid string), return NULL and raise an invalid param error `STUMPLESS_INVALID_PARAM`. You'll need to add this new error type to `include/stumpless/error.h`. Declare your new function `include/stumpless/param.h` and define it in the corresponding source file `src/param.c`.

Add tests for your new function in the existing test suite in `test/function/param.cpp`. Be sure that you test for both success and error cases, particularly escaped characters.

panosfol:
Hello! I would like to take a shot at this!

goatshriek:
Sure thing! I will assign it so that others know it is in work.

panosfol:
@goatshriek should this feature be implemented against the v2.1.0 version?

goatshriek:
[...]

PR #370, opened 2023-09-18
Add stumpless_new_param_from_string function

panosfol:
This function takes a string with a specific format and returns a 'param' struct.
The commit also contains error handling and tests for success and failure cases.

This PR is in regards to issue #360.

### DIFF
--- a/include/private/config/locale/bg-bg.h
+++ b/include/private/config/locale/bg-bg.h
@@ -116,6 +116,10 @@
 "a MULTI_SZ registry value was neither empty nor terminated with two NULL" \
 " characters"
 
+// todo translate
+#  define L10N_INVALID_PARAM_ERROR_MESSAGE \
+"the string is not of the name=\"value\" format"
+
 #  define L10N_INVALID_SEVERITY_ERROR_CODE_TYPE \
 "невалидната тежест"
 
--- a/include/stumpless/error.h
+++ b/include/stumpless/error.h
@@ -126,8 +126,13 @@ extern "C" {
  *
  * @since release v2.1.0
  */\
-  ERROR( STUMPLESS_WINDOWS_FAILURE, 28 )
-
+  ERROR( STUMPLESS_WINDOWS_FAILURE, 28 ) \
+/**
+ * The format of the given string is incorrect.
+ *
+ * @since release v2.2.0
+ */\
+  ERROR( STUMPLESS_INVALID_PARAM_STRING, 29 )
 
 /**
  * An (enum) identifier of the types of errors that might be encountered.
--- a/include/stumpless/param.h
+++ b/include/stumpless/param.h
@@ -329,6 +329,34 @@ struct stumpless_param *
 stumpless_new_param( const char *name, const char *value );
 
 /**
+ * Creates a new param given a string by parsing the string and calling stumpless_new_param.
+ *
+ * **Thread Safety: MT-Safe
+ * This function is thread safe, assuming the the given string is not changed
+ * by other threads during execution.
+ *
+ * **Async Signal Safety: AS-Unsafe heap**
+ * This function is not safe to call from signal handlers since it calls
+ * stumpless_new_param() which has usage of memory management functions.
+ *
+ * **Async Cancel Safety: AC_Unsafe heap**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, since it calls stumpless_new_param() which has usage of
+ * memory management functions.
+ *
+ * @since release v2.2.0
+ *
+ * @param string The string to create the stumpless_param from.
+ *
+ * @return The created param, if no error is encountered. If an error is
+ * encountered, NULL is returned and an error code set appropriately.
+ */
+  
+STUMPLESS_PUBLIC_FUNCTION
+struct stumpless_param *
+stumpless_new_param_from_string( const char *string );
+
+/**
  * Sets the name of the given param.
  *
  * **Thread Safety: MT-Safe**
--- a/src/param.c
+++ b/src/param.c
@@ -143,6 +143,76 @@ stumpless_new_param( const char *name, const char *value ) {
 }
 
 struct stumpless_param *
+stumpless_new_param_from_string( const char *string ) {
+  
+  VALIDATE_ARG_NOT_NULL( string );
+
+  int i;
+  size_t name_len = 0;
+  size_t name_start = 0;
+  size_t value_len = 0;
+  size_t value_start = 0;
+  char *name;
+  char *value;
+  struct stumpless_param *result;
+
+  /* Check that the characters in 'name' are allowed. */
+  for (i = 0; string[i] != '='; i++) {
+    if (!((string[i] >= 'a' && string[i] <= 'z') ||
+	  (string[i] >= 'A' && string[i] <= 'Z') ||
+	  (string[i] >= '0' && string[i] <= '9') ||
+	  (string[i] == '-') ||
+	  (string[i] == '.') ||
+	  (string[i] == '_'))){
+      raise_invalid_param(  );
+      return NULL;
+    }
+  }
+  
+  /* Check if the character after the '=' is '"', else raise an error. */
+  if (string[i + 1] != '"'){
+    raise_invalid_param(  );
+    return NULL;
+  }
+
+  name_len = i;
+  /* We add 1 to the length for the '\0' character. */
+  name = alloc_mem( name_len + 1 );
+  /* We make this addition to skip '=' and '"' characters. */
+  i += 2;
+  value_start = i;
+
+  /* Iterate to the end of the 'value' string. */
+  //for (; string[i] != '\0'; i++){}
+  while (string[i] != '\0'){
+    i++;
+  }
+  /* If the final character isn't '"' we raise an error. */
+  if (string[i - 1] != '"') {
+    raise_invalid_param(  );
+    /* We need to free name here to avoid memory leaks */
+    free_mem( name );
+    return NULL;
+  } else {
+    /* We make this substraction for the 2 '"' characters and for the '=' character. */
+    value_len = i - name_len - 3;
+  }
+  
+  /* We add 1 to the length for the '\0' character. */
+  value = alloc_mem( value_len + 1 );
+  
+  memcpy(name, string + name_start, name_len);
+  name[name_len] = '\0';
+  memcpy(value, string + value_start, value_len);
+  value[value_len] = '\0';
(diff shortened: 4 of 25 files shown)


include/private/config/locale/bg-bg.h
```diff
@@ -280,4 +284,3 @@ L"Stumpless registration of Windows Event Log Source"
 #  define L10N_WSAGETLASTERROR_ERROR_CODE_TYPE \
 "резултатът от WSAGetLastError след неуспешното извикване"
 
-#endif /* __STUMPLESS_PRIVATE_CONFIG_LOCALE_BG_BG_H */
```
goatshriek: I'm not sure why the ending `#endif` statements have been dropped here. I just tested the add_l10n_string script and it does not seem to do this - how did you modify these files?

include/stumpless/error.h
```diff
@@ -126,8 +126,13 @@ extern "C" {
  *
  * @since release v2.1.0
  */\
-  ERROR( STUMPLESS_WINDOWS_FAILURE, 28 )
-
+  ERROR( STUMPLESS_WINDOWS_FAILURE, 28 ) \
+/**
+ * The format of the given string is incorrect.
+ *
+ * @since release v2.1.0
```
goatshriek: This should be 2.2.0, as this will be the next release

include/stumpless/param.h
```diff
@@ -328,6 +328,34 @@ STUMPLESS_PUBLIC_FUNCTION
...
+ * **Async Signal Safety: AS-Unsafe heap**
+ * This function is not safe to call from signal handlers since it calls
+ * stumpless_new_param() which has usage of memory management functions.
+ *
+ * **Async Cancel Safety: AC_Unsafe heap**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, since it calls stumpless_new_param() which has usage of
+ * memory management functions.
+ *
+ * @since 2.1.0
```
goatshriek: This should be 2.2.0 as well.

src/param.c
```diff
@@ -142,6 +142,63 @@ stumpless_new_param( const char *name, const char *value ) {
   return result;
 }
 
+struct stumpless_param *
+stumpless_new_param_from_string( const char *string ) {
+
+  VALIDATE_ARG_NOT_NULL( string );
+
+  int i;
+  int name_len = 0, name_start = 0, value_len = 0, value_start = 0;
```
goatshriek: move these declarations to their own lines, instead of grouping them together.

goatshriek: These should be `size_t`, as they are all used as sizes/indexes, and should not be interpreted as signed in any case.

src/windows/stumpless.def
```diff
@@ -210,15 +210,18 @@ EXPORTS
   stumpless_set_entry_procid                    @195
   stumpless_get_entry_hostname                  @196
   stumpless_set_entry_hostname                  @197
-
+  stumpless_new_param_from_string               @198
```
goatshriek: this should be below the v2.2.0 comment, not above it

src/param.c
```diff
@@ -142,6 +142,63 @@ stumpless_new_param( const char *name, const char *value ) {
...
+
+  /* Check that the characters in 'name' are allowed. */
+  for (i = 0; string[i] != '='; i++) {
```
goatshriek, line 160: what happens if there is no equals sign in the string? it seems like this loop will iterate past the end? Please add a test case for this possibility.

src/param.c
```diff
@@ -142,6 +142,63 @@ stumpless_new_param( const char *name, const char *value ) {
...
+  } else {
+    /* We make this substraction for the 2 '"' characters and for the '=' character. */
+    value_len = i - name_len - 3;
+  }
+  
+  /* Declare an array to hold the 'value' string.
+     We add 1 to the length for the '\0' character. */
+  char value[value_len + 1];
+  
+  strncpy(name, string + name_start, name_len);
```
goatshriek: `memcpy` is a better choice here, as you already know the length. This may also correct the Windows builds that are currently failing.

goatshriek (changes requested):
Thank you very much for putting this together! There are a few changes below for you to make, after which I'll do another final review. Please pay particular attention to fixing the locale and Windows builds that are failing now.

include/private/config/locale/bg-bg.h
panosfol: I did it through the script. Don't know why it dropped them ill check again.

goatshriek: If you are able to reproduce this with the script, please share the version of Ruby and command invocation used and I will see if I can find it.

src/param.c
panosfol: If there is no equal sign, then we iterate until ```string[i]=='\0'``` in which case we will raise an error anyway (since ```\0``` is not an acceptable character for the name) and exit with NULL. Want me to add a clause to end the loop anyway?

goatshriek: As long as you have a test case demonstrating that the case is handled correctly with the correct error, no changes are necessarily needed.

test/function/param.cpp
```diff
@@ -628,4 +628,80 @@ namespace {
...
   }
+
+  TEST( ParamFromStringTest, New ) {
+    struct stumpless_param *param;
+    const char *string = "test-param-name=\"test-param-value\"";
+
+    size_t name_length = strlen("test-param-name");
+    size_t value_length = strlen("test-param-value");
+
+    param = stumpless_new_param_from_string( string );
```
goatshriek: The use of a variable named `string` is confusing the header check tool, which is why you see the static analysis CI failure. It believes it is a reference to the `string` type in C++. In this case, I think it would be better to rename this variable so that there isn't a name re-use, perhaps to something like `param_string`.

test/function/param.cpp
```diff
@@ -628,4 +628,80 @@ namespace {
...
+    const struct stumpless_error *error;
+    
+    param = stumpless_new_param_from_string( NULL );
+
+    EXPECT_ERROR_ID_EQ( STUMPLESS_ARGUMENT_EMPTY );
+
+    stumpless_free_all(  );
+  }
+
+  TEST( ParamFromStringTest, InvalidName ) {
```
goatshriek, line 669: The code coverage run noted that one of the error handling branches is not covered by these tests, where the string ends after the first loop. Please add an invalid string that covers this case; I believe an empty string will do it?

src/param.c
```diff
@@ -142,6 +142,71 @@ stumpless_new_param( const char *name, const char *value ) {
...
+  /* Check if the character after the '=' is '"', else raise an error. */
+  if (string[i + 1] != '"'){
+    raise_invalid_param(  );
+    return NULL;
+  }
+
+  name_len = i;
+  /* Declare an array to hold the 'name' string.
+     We add 1 to the length for the '\0' character. */
+  char name[name_len + 1];
```
goatshriek: The Windows builds appear to be failing on these array declarations (this one and line 200). You'll likely need to change these arrays to be dynamically allocated on the heap (via `alloc_mem` and subsequent `free_mem`). I believe the dynamic size of the array based on a variable, while supported in some compilers like GCC, is not universal.

You could keep the name buffer on the stack if you size it to `STUMPLESS_MAX_PARAM_NAME_LENGTH+1` and do an additional length check when scanning the name portion of the string. The value can theoretically be any size and will therefore need to be dynamic. Windows does support dynamic stack-based buffers via `_malloca` and `_freea`, but using this in a portable manner is probably more work than it's worth for this change.

goatshriek (changes requested):
Thank you very much for addressing the previous issues quickly! There are a few more changes for you below, and some of the locale runs are still failing as well.

Were you able to identify whether the l10n string addition script has a problem in it? I'd like to resolve that if it does have something wrong with it.

panosfol:
> Thank you very much for addressing the previous issues quickly! There are a few more changes for you below, and some of the locale runs are still failing as well.
> 
> Were you able to identify whether the l10n string addition script has a problem in it? I'd like to resolve that if it does have something wrong with it.

It seems there was a problem from my side because I couldn't recreate it, it works perfectly fine now.

src/param.c
panosfol: I think its better I just allocate on the heap through ```alloc_mem```.

panosfol:
@goatshriek Also what should I do about the ```custom_analysis``` fail which is in regards of depricated features?

EDIT: I just saw that these are just warnings and the real problem was about the ```string``` you mentioned so it should be good now.

panosfol:
@goatshriek  it seems there was a bad push from my side ill fix it asap

panosfol:
@goatshriek I've fixed the memory leak, and i fixed the alphabetical order of locales, however im having a little trouble identifying what the problem with codecov is.

goatshriek:
The codecov issue is stating that the lines in question are never run by any of the tests. It looks like adding an empty string didn't actually reach the branch in that conditional (lines 171-172). Is this code reachable? If so, add a test with the scenario that it is intended to handle. If not, then double check that the conditional is correct, and that it is needed. You may simply be able to remove it as dead code that is handled by the if branch above it.

panosfol:
> The codecov issue is stating that the lines in question are never run by any of the tests. It looks like adding an empty string didn't actually reach the branch in that conditional (lines 171-172). Is this code reachable? If so, add a test with the scenario that it is intended to handle. If not, then double check that the conditional is correct, and that it is needed. You may simply be able to remove it as dead code that is handled by the if branch above it.

Yes it seems that the condition above covers the "empty string" case, since ```\0``` is not an acceptable character, therefore we raise an error before we reach the condition ```if (string[i] == '\0')```. I think its ok to remove the second ```if``` condition.
Ill keep the "empty-string" test just to cover all cases.

include/stumpless/param.h
```diff
@@ -328,6 +328,34 @@ STUMPLESS_PUBLIC_FUNCTION
...
+ * **Async Signal Safety: AS-Unsafe heap**
+ * This function is not safe to call from signal handlers since it calls
+ * stumpless_new_param() which has usage of memory management functions.
+ *
+ * **Async Cancel Safety: AC_Unsafe heap**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, since it calls stumpless_new_param() which has usage of
+ * memory management functions.
+ *
+ * @since 2.2.0
```
inquisitev:
```suggestion
 * @since release v2.2.0
```

goatshriek (approved):

goatshriek:
Thanks very much for putting this together and seeing it through to the end!

PR #376, opened 2023-09-28
(272) Added the wide string entry message functions

rmknan:
Added the wide string entry message functions. Changed `entry.c, entry.h, entry.cpp and yml `files


rmknan:
#272

include/stumpless/entry.h
```diff
@@ -1340,6 +1340,39 @@ struct stumpless_entry *
...
+ * This function is not safe to call from signal handlers due to the use of a
+ * non-reentrant lock to coordinate changes and the use of memory management
+ * functions to create the new message and free the old one.
+ *
+ * **Async Cancel Safety: AC-Unsafe lock heap**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, due to the use of a lock that could be left locked as well as
+ * memory management functions.
+ *
+ * @since release v2.1.0
```
goatshriek: change this to the next release, v2.2.0

test/function/entry.cpp
```diff
@@ -2395,6 +2395,164 @@ namespace {
...
+  // TEST( SetMessageWideStrTest, AsciiMessage ) {
+  //   struct stumpless_entry *entry;
+  //   const char *ascii_message;
+  //   const struct stumpless_entry *result;
+  //   const char *new_message;
+
+  //   entry = create_empty_entry(  );
+  //   ASSERT_NOT_NULL( entry );
+
+  //   ascii_message = load_corpus( "cstring/ascii" );
```
goatshriek: make a UTF-16 version of this test corpora named "wstring/ascii"

test/function/entry.cpp
```diff
@@ -2395,6 +2395,164 @@ namespace {
...
+  // TEST( SetMessageWideStrTest, LongAsciiMessage ) {
+  //   struct stumpless_entry *entry;
+  //   const char *long_message;
+  //   const struct stumpless_entry *result;
+  //   const char *new_message;
+
+  //   entry = create_empty_entry(  );
+  //   ASSERT_NOT_NULL( entry );
+
+  //   long_message = load_corpus( "cstring/lorem" );
```
goatshriek: make a UTF-16 version of this test corpora named "wstring/lorem"

test/function/entry.cpp
```diff
@@ -2395,6 +2395,164 @@ namespace {
...
+  // TEST( SetMessageWideStrTest, Utf16Message ) {
+  //   struct stumpless_entry *entry;
+  //   const char *utf16_message;
+  //   const struct stumpless_entry *result;
+  //   const char *new_message;
+
+  //   entry = create_empty_entry(  );
+  //   ASSERT_NOT_NULL( entry );
+
+  //   utf16_message = load_corpus( "cstring/zh-cn" );
```
goatshriek: make a UTF-16 version of this test corpora named "wstring/zh-cn"

goatshriek (changes requested):
Thanks for putting this together! A few minor comments below, mostly around these two points:

The malloc failure test isn't working properly, likely because the failure size is incorrect. Don't forget that wide characters take up more than one byte! EDIT: actually, the errors thrown in failing cases are invalid encoding, the cause of which I'm not sure offhand - you'll have to investigate further.

The test corpora are currently UTF-8 strings. You'll need to make UTF-16 versions of these in a new folder, `test/corpora/wstring`, and change your commented out tests to use these (and of course uncomment them so that they run).

goatshriek:
You'll also need to add your new function to `src/windows/stumpless.def` so that the Windows builds include it.

test/function/entry.cpp
rmknan: What do I do to make them UTF-16, in the program writing` L"test-message"` works to make it `wide str`. I tried saving the UTF-8 files (same content) as UTF-16 but the program does not recognize it as wide string.

goatshriek: The `iconv` utility is useful for this, assuming you have access to it. If not I believe it is possible to do this with Powershell, assuming that you're on a Windows box if not a Linux one.

However, now that I'm thinking about this more you will run into trouble with `wchar_t` not being portable to load from a file, due to potentially different endianness and type sizes between platforms. Creating a portable way to do this is beyond the scope of this change - for now you can simply copy the text out of the load_corpus files and use the `L"string"` format to make a literal in the tests instead.

My apologies for the switch on this.

rmknan:
```
error: incompatible pointer types assigning to 'const wchar_t *' from 'const char *'
    new_message = stumpless_get_entry_message( entry );
                 
```
```
TEST( SetMessageWideStrTest, Utf16Message ) {
    struct stumpless_entry *entry;
    const wchar_t *utf16_message;
    const struct stumpless_entry *result;
    const wchar_t *new_message; // if I make this char then all sorts of error pops up

    entry = create_empty_entry(  );
    ASSERT_NOT_NULL( entry );

    utf16_message = L"没有错误\00";
    ASSERT_NOT_NULL( utf16_message );

    result = stumpless_set_entry_message_str_w( entry, L"没有错误\00" );
    EXPECT_EQ( entry, result );
    EXPECT_NO_ERROR;

    new_message = stumpless_get_entry_message( entry );
    EXPECT_NOT_NULL( new_message );
    EXPECT_NO_ERROR;
    EXPECT_STREQ( utf16_message, new_message );

    delete[] utf16_message;
    free( ( void * ) new_message );
    stumpless_destroy_entry_and_contents( entry );
    stumpless_free_all(  );
  }
```
I reckon there has to be a way to make entry wide char? or is there a way to change something in the function to make it work?

goatshriek: You're absolutely right, and this is another snag I had not considered.

You can solve this in two ways, with the first being much more work: implement the function `stumpless_get_entry_message_w` that will return the message in `wchar` format (and of course add tests for this as well).

The second is to simply convert the string you used to set the message to UTF-8, either as a new literal without the `L` prefix, or using a dynamic conversion function. `stumpless_get_entry_message` will return the function in UTF-8 format, so you just need to convert one of the two to do a comparison.

rmknan: I do not understand the second one, if the message I give is in UTF-8 then `stumpless_set_entry_message_str_w` would not accept it right?

goatshriek: Correct, but the result returned by the get message function will be in UTF-8. So you will need to either convert it to UTF-16 or the expected string to also be in UTF-8.

rmknan: So after the `result` variable is computed I manually declare entry as "没有错误\00" would that do the trick?

goatshriek: You won't need to manually add the null character, but yes that is the string literal that you could use to compare with the result of get message after the `_w` call is done, assuming you used the same characters but in wide form.

rmknan: Its not working out, I get gtest errors and various other errors.

rmknan:
```
 result = stumpless_set_entry_message_str_w( entry, L"没有错误" );
    EXPECT_EQ( entry, result );
    EXPECT_NO_ERROR;

    entry = "没有错误";
    new_message = stumpless_get_entry_message( entry );
    EXPECT_NOT_NULL( new_message );
    EXPECT_NO_ERROR;
    EXPECT_STREQ( utf16_message, new_message );
```
Just to make sure I understood it right.

rmknan: If the first method is the only way to make things work, could you guide me on what changes must be made? If you could describe it as you do when you write the issues, would be very helpful.

goatshriek: Your above code snippet looks mostly right, though you've overwritten the `entry` variable with a string literal before calling `stumpless_get_entry_message`. That will probably cause some errors, though I'm not sure if they're the ones you are encountering. If you still run into problems after this, please provide the errors and your current code and I'll take a look at the specific fault.

rmknan: I thought overwriting the entry variable like I did is what you meant. If that's not it then could you explain how else should entry variable be written?

rmknan:
```
 result = stumpless_set_entry_message_str_w( entry, L"没有错误" );
    EXPECT_EQ( entry, result );
    EXPECT_NO_ERROR;

    // what should come here?
    new_message = stumpless_get_entry_message( entry );
    EXPECT_NOT_NULL( new_message );
    EXPECT_NO_ERROR;
    EXPECT_STREQ( utf16_message, new_message );
```

goatshriek: Try something like this:

```c
const wchar_t *wide_message= L"没有错误";
const char *utf8_message = "没有错误";
const char *new_message;

result = stumpless_set_entry_message_str_w( entry, wide_message );
EXPECT_EQ( entry, result );
EXPECT_NO_ERROR;

new_message = stumpless_get_entry_message( entry );
EXPECT_NOT_NULL( new_message );
EXPECT_NO_ERROR;
EXPECT_STREQ( utf8_message, new_message );
```

rmknan:
```
Expected equality of these values:
  entry
    Which is: 0x562541653640
  result
    Which is: NULL
/home/rmknan/Github/Stump2/stumpless/test/function/entry.cpp:2530: Failure
Value of: stumpless_has_error( )
  Actual: true
Expected: false
an unexpected error occurred: STUMPLESS_INVALID_ENCODING: could not convert the provided wide char string to a multibyte string (errno after the failed call: 84)
/home/rmknan/Github/Stump2/stumpless/test/function/entry.cpp:2535: Failure
Expected equality of these values:
  utf8_message
    Which is: "\xE6\xB2\xA1\xE6\x9C\x89\xE9\x94\x99\xE8\xAF\xAF"
    As Text: "没有错误"
  new_message
    Which is: "fixture message"
<end of output>
```
```
TEST( SetMessageWideStrTest, Utf16Message ) {
    struct stumpless_entry *entry;
    const wchar_t *utf16_message;
    const struct stumpless_entry *result;
    const char *utf8_message = "没有错误";
    const char *new_message;

    entry = create_empty_entry(  );
    ASSERT_NOT_NULL( entry );

    utf16_message = L"没有错误";
    ASSERT_NOT_NULL( utf16_message );
    // ASSERT_NOT_NULL( utf8_message );

    result = stumpless_set_entry_message_str_w( entry, utf16_message );
    EXPECT_EQ( entry, result );
    EXPECT_NO_ERROR;

    new_message = stumpless_get_entry_message( entry );
    EXPECT_NOT_NULL( new_message );
    EXPECT_NO_ERROR;
    EXPECT_STREQ( utf8_message, new_message );

    delete[] utf16_message;
    free( ( void * ) new_message );
    stumpless_destroy_entry_and_contents( entry );
    stumpless_free_all(  );
  }
```

goatshriek: Based on the error, it looks like the test is working properly! Now you'll need to find out why that encoding error is happening, which is likely an issue with the implementation instead of the test.

Keep in mind that the issue may be in the existing wstring code, so be sure to consider that code as well as you work through the issue.

rmknan: Is there a way to debug the code like line by line? I mean if I could place a marker on the wstring code it could lead me to the individual piece of code which is failing. 

I see that the wstring code calls for three different h files. 

Could you also tell me what does these two stand for` HAVE_WINDOWS_H` and` HAVE_WCSRTOMBS_S`?

goatshriek: A standard debugger will let you step through your code line by line, `gdb` being one of the most prominent choices. Many IDEs come with debugger integrations that make this easier to navigate, but that depends on your IDE.

The `HAVE_` symbols describe whether given functions and/or headers are available in the build system. Those two indicate whether `<windows.h>` and the function `wcsrtombs_s` are present, respectively.

rmknan: As you said the `config_copy_wstring_to_cstring` is not working as intended.
Out of the 3 header files being called, the wcsrtomb is the one being used by the program, and its definition is 
```

char *
no_wcsrtombs_s_copy_wstring_to_cstring( const wchar_t *str, int *copy_size );
```
I tried calling the other header files instead of this one but they gave out various errors as well. Currently going through the `stumpless/src/config/no_wcsrtombs_s.c` file.

When I build the program, it shows it as SEGFAULT. Does it mean that I need to focus only on the memory allocation?

```
The following tests FAILED:
	 22 - entry (SEGFAULT)

```

goatshriek: Hmm, I'm not sure what the reason for the segfault is - if you just run it with `gdb` or a similar debugger you should be able to get a backtrace to where the failure occurred. This would be my normal next step.

If you continue to run into roadblocks and don't know where to go next, push what you have and let me know, in which case I'll work out the issues and co-author this change with you. There is also no particular rush on this right now, so if you'd like sole authorship you can keep working through it.

rmknan: Could you tell me how exactly I debug using gdb? I tried to compile the `stumpless/src/config/no_wcsrtombs_s.c` but it gives me an error saying cannot find the` include files`. I  tried to compile that particular file, which I don't think was correct, I was not sure how to go about it.

goatshriek: A detailed tutorial on debugging is beyond the scope of what I can provide. At a high level, make sure that you configure in debug mode (`-DCMAKE_BUILD_TYPE=Debug` provided to the cmake), and then run `gdb` from the source directory. Here is some output from an example debug session, demonstrating how to set a breakpoint in your test, step into library code, and print information at one point in the source.

```
$ gdb ~/build-debug/function-test-entry
GNU gdb (Ubuntu 8.1.1-0ubuntu1) 8.1.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ~/build-debug/function-test-entry...done.
(gdb) b test/function/entry.cpp:97
Breakpoint 1 at 0x2a254: file test/function/entry.cpp, line 97.
(gdb) run
Starting program: ~/build-debug/function-test-entry
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Running main() from ~/build-debug/gtest/src/gtest/googletest/src/gtest_main.cc
[==========] Running 172 tests from 44 test suites.
[----------] Global test environment set-up.
[----------] 92 tests from EntryTest
[ RUN      ] EntryTest.AddElement

Breakpoint 1, (anonymous namespace)::EntryTest_AddElement_Test::TestBody (this=0x84dbd20) at test/function/entry.cpp:97
97          element = stumpless_new_element( "test-new-element" );
(gdb) step
stumpless_new_element (name=0x8083f7a "test-new-element") at src/element.c:478
478     stumpless_new_element( const char *name ) {
(gdb) next
483       VALIDATE_ARG_NOT_NULL( name );
(gdb)
485       if( unlikely( !validate_element_name( name, &name_length ) ) ) {
(gdb)
489       element = alloc_mem( sizeof( *element ) );
(gdb)
490       if( unlikely( !element ) ) {
(gdb) p element
$1 = (struct stumpless_element *) 0x84de0b0
(gdb)
```

rmknan: Based on the debugging of the code, the following code fails and the fail condition is met. (`no_wcsrtombs_s.c`)

```
conversion_result = wcsrtombs( NULL, &str, 0, &state );
  if( conversion_result == -1 ) {
    raise_wide_conversion_failure( errno, L10N_ERRNO_ERROR_CODE_TYPE );
    goto fail;
  }
```
[Stump Debug Sess.zip](https://github.com/goatshriek/stumpless/files/12822959/Stump.Debug.Sess.zip)

I have attached what the debug session goes through if you would like to have a look. Its not able to find several files that are associated with  `wcsrtombs` and therefore is not able to do the conversion from my understanding.

goatshriek: Thank you for providing a debug session to review. The errors regarding missing files are simply reflections of the debugger not having access to standard library files - they are expected if you try to step into a library function like `wcsrtombs`. When you encounter a source line with such a call, often you want to use `next` instead of `step` to advance over it instead of step into it.

I was able to reproduce the error you are encountering locally based on the current state of this pull request. After some quick inspection, I was able to resolve the error by initializing the `mbstate_t` conversion state, a step that was left out by mistake. After adding this, the current tests pass. The new function looks like this:

```c
char *
no_wcsrtombs_s_copy_wstring_to_cstring( const wchar_t *str, int *copy_size ) {
  size_t buffer_size;
  char *buffer;
  mbstate_t state;
  size_t conversion_result;

  memset( &state, 0, sizeof( state ) ); // this is the new line (you'll also need to add an include of string.h)
  conversion_result = wcsrtombs( NULL, &str, 0, &state );
  if( conversion_result == -1 ) {
    raise_wide_conversion_failure( errno, L10N_ERRNO_ERROR_CODE_TYPE );
    goto fail;
  }
  // -- snip --
```

Give this a shot and see if it works for you.

rmknan: Yesterday's commit had the `UTF16 check` commented out. There was some confusion and It seems I had reset the changes to previous commit.

Apologies for that. I have pushed the new commit with the changes now. I tried adding the above piece of code,The build check still says `segfault` and the wide conversion is not happening.

goatshriek: Thank you for the update. The segfault is happening because of the use of  `delete` and `free` on string literals - remove both of those calls from the end of the UTF16 test.

You will also likely need to call `setlocale(LC_ALL, "");` at the beginning of your test so that the conversion does not fail due to a locale issue. This should not be done in the library code itself as it should not modify locale settings, but must be done in order for the test conversion to succeed.

Finally, I would rename the test so that it refers to wide characters instead of UTF16. While debugging I noticed that my system uses UTF32 for wide characters, so using a name referring to UTF16 would be incorrect in some cases.

rmknan: > You will also likely need to call `setlocale(LC_ALL, "");`

So Where would i call it? and i cannot call it in the file entry.cpp file right?

goatshriek: Call it at the beginning of your `TEST` function in `entry.cpp`. You will also need to add the appropriate include, of course.

rmknan:
```
home/rmknan/Github/Stump2/stumpless/test/function/entry.cpp:2535: Failure
Value of: (new_message) == __null
  Actual: true
Expected: false
/home/rmknan/Github/Stump2/stumpless/test/function/entry.cpp:2536: Failure
Value of: stumpless_has_error( )
  Actual: true
Expected: false
an unexpected error occurred: STUMPLESS_MEMORY_ALLOCATION_FAILURE: a memory allocation call failed
/home/rmknan/Github/Stump2/stumpless/test/function/entry.cpp:2537: Failure
Expected equality of these values:
  utf8_message
    Which is: "\xE6\xB2\xA1\xE6\x9C\x89\xE9\x94\x99\xE8\xAF\xAF"
    As Text: "没有错误"
  new_message
    Which is: NULL
```

The `Segfault` is solved but the conversion  still fails.

goatshriek: If you push the current state I will take a look. Note that you also need to merge in the `latest` branch to your local branch, as some conflicts have arisen since you started.

rmknan: Were you able to take a look?

goatshriek: You've left a large portion of `no_wcsrtombs_s_copy_wstring_to_cstring` commented out, which is part of the problem.

You should be seeing compilation warnings regarding passing a size_t pointer into the wstring to cstring function, which takes an int pointer. Correct this error (by changing `new_message_length` to an int) and the problem should resolve (at least it did for me).

rmknan: I was trying something by commenting those out but did not work. The errors have been resolved, did not notice the int `new_message_length`. Thank you, I corrected the other two test the ASCII and Long ASCII as well and it runs successfully.

rmknan:
I changed the stumpless def file, there were some additional text which gave the error. I am not entirely sure about the valgrind test errors.

goatshriek:
The valgrind errors are happening because there is a memory leak in the tests. Be sure that you free the result of `stumpless_get_entry_message` after you're finished with it, as it is dynamically allocated.

rmknan:
I think they should be good to go.

rmknan:
I will look into it again, the tests and build did pass locally.

rmknan:
Hello,

When I try to run Valgrind tests, it does not run and says a file is corrupted as it did in the previous issue. Last time I i solved that by cloning the repo again and making the changes in the new repo as valgrind ran without any problems in the new one.

Now, In the current repo I face the same issue and valgrind does not run. Therefore I tried to reclone the repo again into a separate folder and pasted the changes there. It passes the tests initially and everything runs fine but if I try to make any changes (even if it is commenting ) and re run the checks it says 
```

usr/bin/ld: CMakeFiles/function-test-entry.dir/test/function/entry.cpp.o: in function `(anonymous namespace)::SetMessageWideStrTest_AsciiMessage_Test::TestBody()':
entry.cpp:(.text+0x45dc4): undefined reference to `stumpless_set_entry_message_str_w'
```
The entry message is mentioned in the yml file but even then it says this.

rmknan:
Hello,

When I try to run Valgrind tests, it does not run and says a file is corrupted as it did in the previous issue. Last time I i solved that by cloning the repo again and making the changes in the new repo as valgrind ran without any problems in the new one.

Now, In the current repo I face the same issue and valgrind does not run. Therefore I tried to reclone the repo again into a separate folder and pasted the changes there. It passes the tests initially and everything runs fine but if I try to make any changes (even if it is commenting ) and re run the checks it says

```
/usr/bin/ld: CMakeFiles/function-test-entry.dir/test/function/entry.cpp.o: in function `(anonymous namespace)::SetMessageWideStrTest_AsciiMessage_Test::TestBody()':
entry.cpp:(.text+0x45dc4): undefined reference to `stumpless_set_entry_message_str_w'
/usr/bin/ld: CMakeFiles/function-test-entry.dir/test/function/entry.cpp.o: in function `(anonymous namespace)::SetMessageWideStrTest_LongAsciiMessage_Test::TestBody()':
entry.cpp:(.text+0x467d4): undefined reference to `stumpless_set_entry_message_str_w'
/usr/bin/ld: CMakeFiles/function-test-entry.dir/test/function/entry.cpp.o: in function `(anonymous namespace)::SetMessageWideStrTest_MallocFailureOnMessage_Test::TestBody()':
entry.cpp:(.text+0x47396): undefined reference to `stumpless_set_entry_message_str_w'
/usr/bin/ld: CMakeFiles/function-test-entry.dir/test/function/entry.cpp.o: in function `(anonymous namespace)::SetMessageWideStrTest_NullEntry_Test::TestBody()':
entry.cpp:(.text+0x47b8b): undefined reference to `stumpless_set_entry_message_str_w'
/usr/bin/ld: CMakeFiles/function-test-entry.dir/test/function/entry.cpp.o: in function `(anonymous namespace)::SetMessageWideStrTest_NullMessage_Test::TestBody()':
entry.cpp:(.text+0x483bc): undefined reference to `stumpless_set_entry_message_str_w'
/usr/bin/ld: CMakeFiles/function-test-entry.dir/test/function/entry.cpp.o:entry.cpp:(.text+0x48e0a): more undefined references to `stumpless_set_entry_message_str_w' follow
```



The entry message is mentioned in the yml file but even then it says this.

whereas the current repo's valgrind says this

```
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x23
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x25
### unhandled dwarf2 abbrev form code 0x23
==15962== Valgrind: debuginfo reader: ensure_valid failed:
==15962== Valgrind:   during call to ML_(img_get)
==15962== Valgrind:   request for range [1956341786, +4) exceeds
==15962== Valgrind:   valid image size of 2202296 for image:
==15962== Valgrind:   "/home/rmknan/Github/Stump2/build/function-test-buffer"
==15962== 
==15962== Valgrind: debuginfo reader: Possibly corrupted debuginfo file.
==15962== Valgrind: I can't recover.  Giving up.  Sorry.
==15962== 

```

goatshriek:
It will be some time before I can get to troubleshooting this myself, as there are a few other pieces of the library that I'm spending time on right now. The errors you mentioned above seem like you may need to re-build the test after changing source, try building the `function-test-entry` target and running again.

rmknan:
> The valgrind errors are happening because there is a memory leak in the tests. Be sure that you free the result of `stumpless_get_entry_message` after you're finished with it, as it is dynamically allocated.

`new_message` is the result of `get_entry_message`. So doing` free( ( void * ) new_message );` should solve the valgrind error right?

goatshriek:
Yes, that's likely where the memory leak is coming from.

rmknan:
I did that but valgrind still says the same

rmknan:
Is there anything else I need to look at?

goatshriek:
If you feel that you've exhausted what you're able to resolve, you can push your current state and I'll go through and resolve the remaining errors.

Unfortunately, I don't have the time to go through and create another review detailing the issues since doing a local checkout, troubleshooting, and writeup is quite time consuming. If you do find that you need me to complete this change, I'll start with your current state and open a new PR with everything resolved. You'll have the opportunity to provide a final review of that before it's merged.

Note that this will result in you being a co-author (with me as the other) of the final commit instead of the sole author. If this is important to you, you have lots of time to keep working through this. I certainly don't want you to feel like I've stolen credit for your hard work, so I won't move forward with that course of action unless you let me know that's how you want to proceed, this goes stale as defined in the [inactivity policy](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md#you-still-there), or we approach the next release deadline, which will probably be in December of this year.

rmknan:
I understand and thank you for sharing that. I will work on this and try to come up with a solution. I can not think of any reason for the `valgrind` to fail now so I think I will focus on other issues and take this one slow. If you do get some free time and are able to share any insight please do. 

I honestly don't mind being a co author but I would love to still solve this.

goatshriek:
Incorporated into #412

PR #379, opened 2023-10-10
Added benchmark score for PrivalFromString (#378)

kirubaspace:
This adds benchmark testing for PrivalFromString. This is intended to compare benchmarking score for reference in the future if further optimisations/changes done in the corresponding function.

image


CMakeLists.txt
```diff
@@ -1010,8 +1010,10 @@ add_custom_target(examples
 # performance tests
 add_performance_test(element
   SOURCES test/performance/element.cpp
+  test/performance/priority.cpp
```
goatshriek: Instead of adding these tests to an existing performance test, use the `add_performance_test` function to create a new one named `priority`.

benchmark/src/benchmark/cmake/Modules/FindLLVMAr.cmake
goatshriek, line 1: There are four files included in this PR under the `benchmark` root folder, seemingly accidentally added to git. Remove them from the PR as these should not be tracked in the repository.

test/performance/priority.cpp
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+#include <benchmark/benchmark.h>
+#include <stumpless.h>
+#include "test/helper/memory_counter.hpp"
+
+NEW_MEMORY_COUNTER( prival_from_string )
+
+static void PrivalFromString(benchmark::State& state) {
+  const char *string = "user.info";
```
goatshriek: Variables named `string` can cause confusion with the header checking tool due to a clash with the `string` class provided by C++. Change this variable to a more descriptive name, perhaps something like "priority"?

test/performance/priority.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+  INIT_MEMORY_COUNTER( prival_from_string );
+
+  // Measure the time it takes to call the function.
+  for (auto _ : state) {
+    int result = stumpless_prival_from_string(string);
```
goatshriek: This loop currently benchmarks the function with the input of `user.info`.

At a minimum, there should also be a benchmark that measures the performance of the function when a string containing a number is provided to it. It may make more sense for this to be a separate benchmark in the same `priority.cpp` file.

Also consider how you could benchmark more than just one of the possible input strings. Building an array of possible inputs before starting the `for` loop is one way to do this. It doesn't necessarily need to be comprehensive, but it would be nice to exercise a few more possible inputs.

goatshriek (changes requested):
Thanks very much for putting this together! I've included a few comments for you to address, after which I will run the CI tests.

test/performance/priority.cpp
jagw-mobica: Yes, I suggest to use this list of input strings: ["191", "user.emerg", "uucp.err", "local7.debug"].
Every string in this list has different run time so benchmark would more valuable.

kirubaspace: Thanks for your input. 

Updated benchmark scores 

image

CMakeLists.txt
kirubaspace: hi @goatshriek this one I could not figure out why its not working for me :( 

I've had the below change 

# performance tests
add_performance_test(element
  SOURCES test/performance/element.cpp
)

add_performance_test(priority
  SOURCES test/performance/priority.cpp
)

but when I run benchmark even after 'make clean' its not picking up the priority.cpp file. Could you please let me know what I'm missing here. Hence, as of now, added that priority.cpp along with the element tag

goatshriek: I'm not sure why that would be, the usage above should be all that is needed. Sometimes I have seen issues where changing the `CMakeLists.txt` file causes problems in an existing build tree. Try making the change above and then doing the configure and build in a clean directory.

kirubaspace: ok done

goatshriek (changes requested):
Everything looks good now! The final thing for you to correct is simply add an include for `<string>` to `priority.cpp` (due to the usage of `std::string`), which should satisfy the include check in the Static Analysis workflow.

goatshriek (approved):

Issue #358, opened 2023-08-24
expand facility from string


goatshriek:
The Stumpless [Rust crate and CLI tool](https://github.com/goatshriek/stumpless-logger) creates targets and entries to do whatever logging needed. In order to provide CLI support, many of these tasks need to use the strings provided in the command to set up stumpless structures. One of these is the prival used for the severity and facility values of the logged entry. To support this, the existing `stumpless_get_facility_enum` function needs to be enhanced to support the same strings that [standard logging tools](https://man7.org/linux/man-pages/man1/logger.1.html) do.

Currently the facility enum from string function only supports strings that exactly match the name of the symbol for a facility level, such as "STUMPLESS_FACILITY_AUTH". It needs to be updated to support the following strings, as well as not being case sensitive.

```
auth
authpriv
cron
daemon
ftp
kern
lpr
mail
news
syslog
user
uucp
local0
local1
local2
local3
local4
local5
local6
local7
security
```

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, review the existing implementation of `stumpless_get_facility_enum` in `src/facility.c`. Once you understand how the current implementation works, start adding the new functionality. Note that the list of new strings to accept mostly matches the end of the current symbol names without the `STUMPLESS_FACILITY_` prefix - you can likely use this fact to simplify your implementation.

Add tests for your new functionality in the existing test module `test/function/facility.cpp`, as separate test cases from the existing ones for this function. Be sure that you test for upper and lower case versions of the strings (both should be accepted), invalid strings, empty strings, and NULL strings at a minimum.

EDIT: this issue was originally duplicative of #237 and has been updated to specify the next phase of intended functionality.

jagw-mobica:
Hi, I want to do this task, but I'm not sure if I understand it correctly.
There is already created function  `stumpless_get_facility_enum( const char *facility_string )`.
What is difference between this expected to do `stumpless_facility_from_string` and  `stumpless_get_facility_enum( const char *facility_string )`?
Can you provide an usage example for it?

goatshriek:
You're absolutely correct, I apologize for that. I had forgotten this was already done and missed the function when I did a quick check before making the issue.

I've updated this one to refer to the next thing that needs to be done with this function - please give the revised description a look and see if it's still something that interests you.

jagw-mobica:
[...]

PR #380, opened 2023-10-11
expand facility from string (#358)

jagw-mobica:
Modify stumpless_get_facility_enum that the given string matches the end of the current symbol names without the STUMPLESS_FACILITY_ prefix. Exception: 'security' can be treated as identical to 'auth' and is returning STUMPLESS_FACILITY_AUTH.
This is intended to support CLI and other string-based applications.


jagw-mobica:
I had to correct the logic in condition (file: priority.c). Now it should works.
Additionally I added test cases for this condition.

test/function/facility.cpp
```diff
@@ -53,11 +57,59 @@ namespace {
...
+      result = stumpless_get_facility_enum( #STRING + 19 ); \
       EXPECT_EQ( result, ENUM );
     STUMPLESS_FOREACH_FACILITY( CHECK_FACILITY_ENUM )
   }
 
+  TEST( GetFacilityEnum, LowercaseFacility ) {
+    int result;
+
+    result = stumpless_get_facility_enum( "user" );
+    EXPECT_EQ( result, 8 );
```
goatshriek: For each of these comparisons, use the enum symbol (e.g. `STUMPLESS_FACILITY_USER`) instead of the hardcoded number.

goatshriek (changes requested):
Thank you very much for putting this together! Just one adjustment for you to make in order for the tests to be more readable.

test/function/facility.cpp
jagw-mobica: Good point, done.

src/priority.c
```diff
@@ -43,8 +38,7 @@ stumpless_prival_from_string( const char *string ) {
   size_t len;
   size_t slen;
   size_t pre_len;
-  const char pre_facility[] = "STUMPLESS_FACILITY_";
-  const char pre_severity[] = "STUMPLESS_SEVERITY_";
+//  const char pre_severity[] = "STUMPLESS_SEVERITY_";
```
goatshriek: remove commented out code

src/priority.c
```diff
@@ -94,18 +84,24 @@ stumpless_prival_from_string( const char *string ) {
 
   // Calculate the severity length
   len = slen - ++len;
-  pre_len = sizeof(pre_severity) - 1;
+//  pre_len = sizeof(pre_severity) - 1;
```
goatshriek: remove commented out code

test/function/priority.cpp
```diff
@@ -35,6 +35,20 @@ namespace {
     EXPECT_EQ( result, 119 );
   }
 
+  TEST( GetPriorityValue, NumTooHighPriority ) {
+    int result;
+
+    result = stumpless_prival_from_string( "192" );
+    EXPECT_EQ( result, -1 );
```
goatshriek, line 43: Also call `EXPECT_ERROR_ID_EQ` after these and make sure that the error is set to the appropriate type.

test/function/severity.cpp
```diff
@@ -53,11 +57,52 @@ namespace {
...
+      result = stumpless_get_severity_enum( #STRING + 19 ); \
       EXPECT_EQ( result, ENUM );
     STUMPLESS_FOREACH_SEVERITY( CHECK_SEVERITY_ENUM )
   }
 
+  TEST( GetSeverityEnum, LowercaseValidSeverity ) {
+    int result;
+
+    result = stumpless_get_severity_enum( "emerg" );
+    EXPECT_EQ( result, STUMPLESS_SEVERITY_EMERG );
```
goatshriek, line 69: include `EXPECT_NO_ERROR` after each of these successful calls to make sure that an error was not left around by mistake.

src/priority.c
```diff
@@ -94,18 +84,24 @@ stumpless_prival_from_string( const char *string ) {
...
 
-  param = alloc_mem( len + 1 + pre_len );
+  // Copy the severity substring to the param buffer
+  param = copy_cstring_length( ++period, len );
   if( !param ) {
     return -1;
   }
+
+//  param = alloc_mem( len + 1 + pre_len );
```
goatshriek: remove these blocks of commented code as well

goatshriek (changes requested):
Thank you for putting this together, it is mostly ready! Only a few adjustments for you to make.

src/priority.c
jagw-mobica: done

src/priority.c
jagw-mobica: done

test/function/priority.cpp
jagw-mobica: done

test/function/severity.cpp
jagw-mobica: done

src/priority.c
jagw-mobica: done

goatshriek (approved):

PR #384, opened 2023-10-16
Severity from buffer

### LINKED ISSUE #383: add severity from buffer
While a function already exists to get a severity from a string, `stumpless_get_severity_enum`, it requires a NULL-terminated string. This complicates some other functions, in some cases requiring them to copy a portion of a string in order to NULL-terminate it. This change is to add a new function `stumpless_get_severity_enum_from_buffer` which uses a buffer and a length instead of a NULL-terminated string.

## General Approach

First, read [this](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions) section of the development documentation on adding new functions. Once you understand how this process works, start defining the new function.

Review the existing implementation of `stumpless_get_severity_enum` in `src/severity.c`. Once you understand how the current implementation works, start adding the new functionality to this same `src/severity.c` file and the associated header.

Add tests for your new functionality in the existing test module `test/function/severity.cpp`, based on the tests already there.

kirubaspace:
This fixes https://github.com/goatshriek/stumpless/issues/383 


kirubaspace:
@goatshriek Could  you please review when you get time

goatshriek:
There are merge conflicts that must be resolved before this will be reviewed. It looks like you will need to do a fresh pull from `latest`, as the functionality has changed a bit since this branch appears to have started.

goatshriek (changes requested):
Please review the current `stumpless_get_severity_enum` functionality and tests - it has been expanded recently. You'll need to make sure these newer features are preserved.

Rather than simply copying the functionality directly, it will make more sense to move the functionality to your function, and have the previous function call your new `_from_buffer` one so that code is not duplicated.

kirubaspace:
> Please review the current `stumpless_get_severity_enum` functionality and tests - it has been expanded recently. You'll need to make sure these newer features are preserved.
> 
> Rather than simply copying the functionality directly, it will make more sense to move the functionality to your function, and have the previous function call your new `_from_buffer` one so that code is not duplicated.

Moved the functionality to new function and updated the unit test. Please let me know if this is fine or any changes required

test/function/severity.cpp
```diff
@@ -122,4 +122,24 @@ namespace {
...
+  }
+
+  TEST( GetSeverityEnumFromBuffer, InvalidMemSeverity ) {
+    int result;
+    void * (*set_malloc_result)(size_t);
+    set_malloc_result = stumpless_set_malloc( MALLOC_FAIL );
+    ASSERT_NOT_NULL( set_malloc_result );
+
+    result = stumpless_get_severity_enum_from_buffer( "info", sizeof("info") );
+    EXPECT_EQ( result, -1 );
```
goatshriek, line 140: There should also be an `EXPECT_ERROR_ID_EQ` check here to make sure that the error code is what is expected in this case. It looks like that's also missing in the test that you based this one on, so feel free to add it there as well.

goatshriek (changes requested):
Thanks for putting this together and getting it updated so quickly! Just a few minor tweaks for you to make.

To resolve the Windows errors, you'll need to add your new function to `src/windows/stumpless.def`, which will expose the call to linkers.

goatshriek (approved):

Issue #388, opened 2023-11-01
implement turkish (tr-TR) localization


goatshriek:
There is a detailed guide on how to go about adding a new language in the [localization documentation](https://github.com/goatshriek/stumpless/blob/latest/docs/localization.md).

If you'd like a concrete example of such a change, check out commit 5529e077eb75c6e79d5634e9d7fc708828b78f89, which shows how to update the various build and header files outlined in the documentation.

OsmanBuyuksar:
Hi, i added localization for turkish but is there a way for me to test it before making a pull request?

goatshriek:
The most straightforward way to test is to build similarly to the way the [locale tests](https://github.com/goatshriek/stumpless/blob/latest/.github/workflows/locale.yml) do. Set the `LANG` environment variable appropriately, and then run through a build and make sure that your localization is chosen, most easily by checking the setting in `include/stumpless/config.h` in the build directory. If you want to make absolutely sure, you can temporarily add a [`stumpless_perror`](https://goatshriek.github.io/stumpless/docs/c/v2-1-0/error_8h.html#a26c67184a8490284024dc893a1cf2a53) call to one of the empty argument tests and make sure that the correct localized string is printed instead of the default English one.

OsmanBuyuksar:
Thanks for the help, i ran a build using the command `LANG=tr_TR.UTF-8 cmake ../stumpless_tr_translate/stumpless/` and the compile was succesfull. The pull request is: https://github.com/goatshriek/stumpless/pull/389

PR #389, opened 2023-11-01
Added the turkish localization

OsmanBuyuksar:
Solution for this issue: https://github.com/goatshriek/stumpless/issues/388#issue-1971501202

Implemented turkish localization as described [here](https://github.com/goatshriek/stumpless/blob/latest/docs/localization.md)


CMakeLists.txt
```diff
@@ -210,6 +210,9 @@ elseif(LOCALE MATCHES "^da[_-]")
 elseif(LOCALE MATCHES "^he[_-]")
   set(STUMPLESS_LANGUAGE "he-IL")
   set(USE_LOCALE_HE_IL TRUE)
+elseif(LOCALE MATCHES "^tr[_-]")
+  set(STUMPLESS_LANGUAGE "tr-TR")
+  set(USE_LOCALE_HE_IL TRUE)
```
goatshriek: This should instead be `USE_LOCALE_TR_TR`.

goatshriek (changes requested):
Thank you very much for contributing this! Just two changes to make, the first listed below. The second is to add a new section to the `.github/workflows/locale.yml` for the new translation. Commit 5529e077eb75c6e79d5634e9d7fc708828b78f89 has a good example of how to make this change if you need something to look at for comparison.

OsmanBuyuksar:
Sry about that, i have 2 repos to keep the contributing repo clean and looks like i made some of the changes to other repo :D. Anyways i fixed locale in cmakelists and also added the tr section to locale.yml file. Let me know if there is anything else.

goatshriek:
Looks like there's just one more thing to check - at least one line is longer than 80 characters. Once you've broken that up this should be all good to go.

OsmanBuyuksar:
There shouldn't be any line longer than 80 characters now, is there anything else?

Issue #371, opened 2023-09-19
document private functions


goatshriek:
In the early stages of development of stumpless, documentation of private functions was not always done, as these are not exposed to library users. However, as the library has grown and more contributors have participated, the need for documentation of private routines is becoming more real.

This is a general issue to track documentation of the private functions, and is open to anyone who wishes to contribute without writing new code, but rather simply wants to read existing code and understand it. Go through the header files in `include/private` and add a doxygen-style comment for functions that currently don't have one. Be particularly careful to cover the [POSIX safety](https://www.gnu.org/software/libc/manual/html_node/POSIX-Safety-Concepts.html) of the function, which outline the thread, signal, and asynchronous cancellation safety of the function. Look to other functions for examples of the formatting - all functions in the `include/stumpless` headers are documented already. You can find the implementation of all functions in the file in the `src` directory corresponding to the header file.

In order for a pull request to be accepted regarding this issue, please document at a minimum three (3) functions, but the more the better, of course!

vanshiz:
Please assign me this.

goatshriek:
Hi @vanshiz, thank you for your interest and for reaching out!

This issue is a general placeholder, so it won't be assigned to anyone specifically. If you'd like an issue assigned to you so others can see what you're working on, feel free to open a separate one listing the specific functions you intend to cover.

It's also completely acceptable to open a PR without your own issue, if opening a new one feels like too much trouble.

blastoncrush:
Hello ! I would like to know if this issue is still relevant, and contribute to it if it is the case :)

goatshriek:
Absolutely, please feel free to look around and find some private functions to understand and document!

ritesh006:
HI @goatshriek this issue is closed now. I think Tere are another function available for doc. Could you please reopen the this issu

goatshriek:
My apologies, it was accidentally closed in another PR.

PR #390, opened 2023-11-02
Private folder doc

OsmanBuyuksar:
Issue: https://github.com/goatshriek/stumpless/issues/371#issue-1902709848

Documented the first 5 functions in include/private/entry.h

I am fairly new to checking for asynchronous safety so it would be nice if someone can review and give feedback.


include/private/entry.h
```diff
@@ -27,24 +27,137 @@
...
+ *
+ * **Async Signal Safety: AS-Unsafe**
+ * This function is not safe to call from signal handlers due to destroying
+ * resources(entry cache) that might be used within a function.
+ *
+ * **Async Cancel Safety: AC-Unsafe**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, due to leaving memory in an undefined state.
+ *
+ * @since release v1.6.0
```
goatshriek: Have you looked up when this function was added in the commit history? I'm not 100% sure when it came into existence myself. If not, there's no need to do that extra work: you can simply remove the `@since` tag. The tag is only mandatory for new functions moving forward. This of course goes for all of the `@since` tags here.

include/private/entry.h
```diff
@@ -27,24 +27,137 @@
...
 void
 entry_free_all( void );
 
+/**
+ * Gets the priority value from facility and severity parameters.
+ * 
+ * Priority value is calculated by left shifting the facility value by 3 
+ * and adding it to the severity value which is according to the 
+ * RFC 5424 Section 6.2.1. The shift operation is done prior to get_prival()
+ * function with macros in "facility.h"
```
goatshriek, line 55: This paragraph may be a bit confusing, as the facility `_VALUE` definitions are already shifted, and do not need to be shifted again. Since this is already taken care of elsewhere, it's probably best to simplify this paragraph to say something along the lines of "priority is calculated using the facility and severity values in accordance with RFC 5424 Section 6.2.1"

include/private/entry.h
```diff
@@ -27,24 +27,137 @@
...
 
+/**
+ * Locks the mutex within the entry.
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe.
+ *
+ * **Async Signal Safety: AS-Unsafe **
+ * This function is not safe to call from signal handlers due to the use of
+ * function 'pthread_mutex_lock()' from <pthread.h> that is not guarenteed to
```
goatshriek: Some builds, for example Windows, don't use the `pthread.h` functions, but instead use the Windows mutex libraries. Simplify this sentence to simply mention "mutexes" and leave out the implementation detail.

include/private/entry.h
```diff
@@ -27,24 +27,137 @@
...
+ * @param index The index of element within the entry.
+ *
+ * @return The element if no error is encountered. If an error is
+ * encountered, then NULL is returned and an error code is set appropriately.
+ */
 struct stumpless_element *
 locked_get_element_by_index( const struct stumpless_entry *entry,
                              size_t index );
 
+
```
goatshriek: remove this extra newline

include/private/entry.h
```diff
@@ -27,24 +27,137 @@
...
+ * **Thread Safety: MT-Unsafe**
+ * This function is not thread safe as it doesn't use any synchronization 
+ * or locking mechanisms while accessing shared resources.
+ *
+ * **Async Signal Safety: AS-Safe**
+ * This function is safe to call from signal handlers.
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously 
+ * canceled.
```
goatshriek: This function is unfortunately AS and AC-unsafe for the same reasons listed in `include/private/error.h` for `raise_error`.

There is some nuance to this and possibly the reasons in on `raise_error` could be updated, and perhaps a case made for `AC-Safe` currently, but for now you'll need to match the `raise_error` safety to be consistent.

goatshriek (changes requested):
Thanks very much for putting this together, and for your patience in waiting for me to get around to reviewing it. There are some comments here for you to apply, after which this will be good to go!

include/private/entry.h
OsmanBuyuksar: Okay, i wasn't sure about those anyway cause it wasn't clear in the commit history so i am removing those tags in the functions that are not moving forward

include/private/entry.h
OsmanBuyuksar: I was a bit confused about this function as i didn't understand why it was calculating priority value that way so i took notes about the function inside the documentation and since i was confused at first i left it there in case someone wonders the same thing that i thought but it can be a bit confusing so i will shorten it as you said.

include/private/entry.h
OsmanBuyuksar: Hmm this is because this uses raise_index_out_of_bounds() right? are there other things too btw? since i didn't checked for async safety before i sometimes doubt about what to write.

goatshriek: That's correct, it's because of the index error, which in turn calls the generic error raising function. I'll need to make the error handling safer at some point, but for now it's marked as unsafe for signals or async cancellation.

If you're asking whether I reviewed your safety determinations for the other functions as well, I did and they looked right to me!

goatshriek:
This will be good to go once the other `@since` tags are removed!

include/private/entry.h
OsmanBuyuksar: Thanks for the feedback i appreciate it :)

OsmanBuyuksar:
My bad i thought it was only the old one, but i removed other since tags now.

goatshriek (approved):

Issue #362, opened 2023-08-26
add benchmark for stumpless_param_to_string


goatshriek:
As the param to string functionality is adjusted to more closely match the formats of RFC 5424 and existing CLI tools (see #361), there is an opportunity to re-use the underlying functionality in the formatting of messages as well. This will make the performance of this function critical, as it will be used for every message formatting call where structured data including a param is present. Adding a benchmark test for `stumpless_param_to_string` will give future changes a way to evaluate what the impact to performance is.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, read the [documentation](https://github.com/goatshriek/stumpless/blob/latest/docs/benchmark.md) on benchmarking in stumpless. This doc walks through the general philosophy behind benchmark tests and how they are used in this project.

Once you understand these concepts, you'll be ready to add a new benchmark in `test/performance/param.cpp` for `stumpless_param_to_string`. Make sure that your new benchmark is included in the results of the `bench` build target and when the param benchmark is run on its own using the `run-performance-test-param` target.

iligeoili:
Hey i would like to take on this issue if its still available.

goatshriek:
Sounds good! I will assign it to you so that others know it is in work.

alessandro-murtas:
Hi, since it seems nobody is working on it, could i work on it?

goatshriek:
Sure! Feel free to open a pull request as soon as you have something ready. I have not seen anything from @iligeoili to indicate that they have made progress on it.

PR #392, opened 2023-12-02
(#362) Added stumpless_param_to_string benchmark

alessandro-murtas:
Added the benchmark for stumpless_param_to_string (Issue #362).

### DIFF
--- a/test/performance/param.cpp
+++ b/test/performance/param.cpp
@@ -25,6 +25,7 @@ NEW_MEMORY_COUNTER( load_param )
 NEW_MEMORY_COUNTER( new_param )
 NEW_MEMORY_COUNTER( set_param_name )
 NEW_MEMORY_COUNTER( from_string )
+NEW_MEMORY_COUNTER( param_to_string )
 
 static void FromString(benchmark::State& state) {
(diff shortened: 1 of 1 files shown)


test/performance/param.cpp
```diff
@@ -130,8 +131,30 @@ static void SetParamName(benchmark::State& state){
...
+static void ParamToString(benchmark::State& state){
+  struct stumpless_param *param;
+  const char *result;
+
+  INIT_MEMORY_COUNTER( param_to_string );
+
+  param = stumpless_new_param( "new-param-name", "new-param-value" );
+
+  for(auto _ : state){
+    result = stumpless_param_to_string( param );
```
goatshriek, line 143: You'll need to free the result of this call so that the memory doesn't explode during the test. Check out `test/performance/version.cpp` for an example, which does something similar.

goatshriek (changes requested):
Thanks for putting this together so quickly! Just one quick change to make below, and this will be good to go!

test/performance/param.cpp
alessandro-murtas: Done! 
Hope it is what you meant!

goatshriek (approved):

PR #396, opened 2023-12-19
Added stumpless_element_to_string benchmark #375

Miroslav263:


goatshriek:
Thanks very much for putting this together! It looks great!

PR #401, opened 2024-01-18
Fixed unsequenced modification warning.

EralpCelebi:
Increments 'len' before assignment to avoid warnings regarding unsequenced modification.


goatshriek (approved):

goatshriek:
Thanks very much for putting this together so quickly!

PR #408, opened 2024-02-23
Adding Sinhala and Japanese localizations

### LINKED ISSUE #283: implement sinhala (si-LK) localization
There is a detailed guide on how to go about adding a new language in the [localization documentation](https://github.com/goatshriek/stumpless/blob/latest/docs/localization.md).

If you'd like a concrete example of such a change, check out commit 5529e077eb75c6e79d5634e9d7fc708828b78f89, which shows how to update the various build and header files outlined in the documentation.

fluidskateboards:
This change resolves #283 and #301 by adding Sinhala and Japanese localizations.


goatshriek (changes requested):
Thank you very much for putting together this update with not just one but _two_ new locales! I believe the only fix needed is to add the `L10N_NETWORK_CLOSED_ERROR_MESSAGE` definition to the `si-lk.h` header, which is causing the two test build failures.

fluidskateboards:
I added the missing header, thanks for reviewing it. I didn't know how to review the issue myself, but now I see the error message with an undeclared header.

goatshriek (approved):

goatshriek:
Thanks once again for putting this together, and for responding quickly to the requested changes! It is greatly appreciated!

PR #412, opened 2024-03-30
add entry message wide string support

### LINKED ISSUE #272: create wide string set entry message functions

goatshriek:
Implements `stumpless_set_entry_message_str_w` and fixes some wide string functionality as well.

This is a continuation of #376, which implemented most of this functionality but had a few problems remaining, and has since gone stale.

This fixes #272, which can be referenced for more information.

### DIFF
--- a/src/entry.c
+++ b/src/entry.c
@@ -887,6 +887,7 @@ stumpless_set_entry_message_str_w( struct stumpless_entry *entry,
     if( !new_message ){
       return NULL;
     }
+    new_message_size--; // leave off the NULL character
 
   } else {
     new_message = NULL;
--- a/src/entry.c
+++ b/src/entry.c
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /*
- * Copyright 2018-2022 Joel E. Anderson
+ * Copyright 2018-2024 Joel E. Anderson
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/entry.c
+++ b/src/entry.c
@@ -33,6 +33,7 @@
 #include "private/config/wrapper/getpid.h"
 #include "private/config/wrapper/thread_safety.h"
 #include "private/config/wrapper/wel.h"
+#include "private/config/wrapper/wstring.h"
 #include "private/deprecate.h"
 #include "private/element.h"
 #include "private/entry.h"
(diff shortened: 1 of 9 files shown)


goatshriek:
@rmknan, this pull is the completion of your previous one. Please feel free to provide a review of it, particularly since it is almost entirely derived from your work. I'll leave it open for a week to give you some time to comment before merging it in.

src/entry.c
```diff
@@ -887,6 +887,7 @@ stumpless_set_entry_message_str_w( struct stumpless_entry *entry,
     if( !new_message ){
       return NULL;
     }
+    new_message_size--; // leave off the NULL character
```
rmknan, line 890: Yes, I am going through the commits. Could you tell me about this line ?

goatshriek: The wide string conversion functions return a size that includes the NULL terminator, but the entry `message_length` member shouldn't include the NULL terminator in it. Stumpless currently assumes that this string is NULL-terminated though, and specifically the `stumpless_get_entry_message` function does when copying out the message to return. If the NULL terminator is left in on this length, the get entry function will try to read an extra byte at the end because it expects a NULL terminator to be there, which is invalid (and causes valgrind errors).

It's a confusing convention, and the next major version (v3.0) will drop the NULL terminators completely for this reason, but for now this line is necessary to keep everything consistent.

Issue #245, opened 2022-10-03
prival to string function


goatshriek:
Entries have both a facility and a severity, combined into a value called the prival. While the individual portions can be converted to strings already, the prival itself does not currently have a function to do this. This will be a necessary function to create in order to easily make an entry into a string. Your task is to implement this function.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, read [this](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions) section of the development documentation on adding new functions. Once you understand how this process works, start defining the new function.

You'll need to create a new header file for this function, `include/stumpless/prival.h`. You can copy `include/stumpless/severity.h` as a template, and alter it as needed. Once this is done, add a new function with the following signature:

```c
STUMPLESS_PUBLIC_FUNCTION
const char *
stumpless_get_prival_string( int prival );
```

Next, create a matching file `src/prival.c` to hold your implementation. You'll need to update the `CMakeLists.txt` file to include the new file in the build. Use the severity and facility to string functions to build the prival string. Use ` | ` to separate the severity and the facility strings, putting the severity first.

Finally, you'll need to also create a new test suite to hold the tests for your function. Once again, you can copy the `test/function/severity.cpp` test suite to hold your new test. And of course, you'll need to add this test suite to the CMake build file as well.

mk722:
Hello, can you assign this issue to me please?

mk722:
Sorry, I haven't started this issue yet, I had a lot in work, but I'll do it this week.

goatshriek:
No worries, thanks very much for the update!

mk722:
Hello, I wanted just to let you know, that I caught covid this week and I can't program at the moment because of it. Are you in a hurry with this issue? I suppose, that for the next week, if I feel better, I would have already finished it.

goatshriek:
No worries at all, `good first issue`s are specifically not time critical to give newcomers time to work on them without feeling rushed. As long as you keep us informed that you are still working on this, you can take as long as you need to.

We wish you a speedy recovery!

mk722:
Hello, I just want to let you know that I am still working on this issue, I just have a more difficult course of covid, so it will take a while. 
Thank you for understanding.

uFarad:
Please see PR #413

PR #413, opened 2024-04-04
added prival to string function

uFarad:
Added function to convert the prival integer to a string in response to issue #245 

@mk722 did great work on this previously, but seems to be inactive now after a bout with COVID. I hope they had a speedy recovery. 

I took over this change from where they left off. Please let me know if anything needs to be changed or if I am not following proper procedures. I hope to contribute to more open source projects like this one in the future.


goatshriek:
Note that the continuous integration tests cannot run until the merge conflicts are resolved.

uFarad:
It successfully compiles and passes tests locally. It should be good to go.

goatshriek:
Thanks very much for getting this together! You can disregard the codecov upload failures, those are due to a separate issue I am working on addressing. I will do a more thorough review of the change in the next few days.

include/private/prival.h
```diff
@@ -0,0 +1,42 @@
...
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled.
+ *
+ * @param prival int to get the string from.
+ *
+ * @return The string representation of the given prival.
+ */
+int
+stumpless_get_prival_string( int prival );
```
goatshriek: Is this function used anywhere? It seems like it is duplicative of the `stumpless/prival.h` function. This header may not be needed at all.

include/stumpless/prival.h
```diff
@@ -0,0 +1,62 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** @file
+ * Function for converting prival from int to string
+ */
+
+#ifndef __STUMPLESS_PRIVAL_STRING
```
goatshriek: the header guard should match the others in the project: this one should be `__STUMPLESS_PRIVAL_H`

include/stumpless/prival.h
```diff
@@ -0,0 +1,62 @@
...
+#    include <syslog.h>
+#  endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+/**
+ * Gets the string representation of the given prival.
+ *
+ * This is a string literal that should not be modified or freed by the caller.
```
goatshriek: The implementation uses `alloc_mem`, which means that this statement is incorrect: the result MUST be freed by the caller to avoid a memory leak.

include/stumpless/prival.h
```diff
@@ -0,0 +1,62 @@
...
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe.
+ *
+ * **Async Signal Safety: AS-Safe**
+ * This function is safe to call from signal handlers.
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled.
+ *
```
goatshriek, line 49: add a `@since` tag for this function (this will be in release 2.2.0)

include/stumpless/prival.h
```diff
@@ -0,0 +1,62 @@
...
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe.
+ *
+ * **Async Signal Safety: AS-Safe**
+ * This function is safe to call from signal handlers.
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled.
```
goatshriek: The use of heap allocation (via `alloc_mem`) likely makes this function neither AS-Safe or AC-Safe. Check the documentation for other functions to see how this is documented.

CMakeLists.txt
```diff
@@ -1025,7 +1031,6 @@ add_custom_target(check
   DEPENDS ${STUMPLESS_FUNCTION_TESTS}
 )
 
-
```
goatshriek: don't remove this extra newline which acts as a section separator

goatshriek (changes requested):
See the requested changes for the first round of revisions.

Also, make sure that you run the header check tool and resolve any header errors. The static analysis CI workflow is currently failing on header checks.

include/private/prival.h
uFarad: The redundant header has been removed. It was indeed not needed.

include/stumpless/prival.h
uFarad: I've fixed the header guard to match the rest of the project.

include/stumpless/prival.h
uFarad: My mistake. I used another file as a template and did not update that info.

include/stumpless/prival.h
uFarad: Tag added!

include/stumpless/prival.h
uFarad: Ditto here. I used another file as a template and did not update that info. I've used other functions as an example of how to document this.

CMakeLists.txt
uFarad: I've re-added this newline.

goatshriek:
Thank you once again for working through this! My apologies for the many confusing CI failures - you had the misfortune of committing during a time of codecov breakage and also discovering a bug in the header checking tool.

uFarad:
It was my pleasure. Thank you for creating an amazing project that is friendly to those new to OSS.

Issue #415, opened 2024-04-07
implement korean (ko-kr) localization


codehooni:
I want to localize Korean to this open source

PR #419, opened 2024-05-26
add ko-kr.h

codehooni:
I sent you a Korean translation full request. Please check it (#415)

###Added files
- include/private/config/locale/ko-kr.h (Korean translation file)

###Files with added content
- include/private/config/locale/wrapper.h
    `# elif defined USE_LOCALE_KO_KR
      #iclude "private/config/locale/ko-kr.h＂`
- README.md
`:ko:` add


goatshriek:
Thanks very much for putting this together!

There is a code conflict due to some recent changes that have been made in header locations, and some other localization-related updates as well. For the current pull request, I believe you just need to account for the `locale/wrapper.h` header being moved to `wrapper/locale.h`, and I can deal with the other changes after this is merged.

PR #424, opened 2024-06-09
Update README.md

codehooni:
There was one mistake I made in the contribution process. The Korean flag was not properly displayed in the README.md file. :kr: was incorrectly marked as :ko: Please change it again. I'm sorry for the inconvenience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Stumpless has lots of features that make logging in C fast and easy:
  * can be adjusted or removed during compilation for zero runtime cost
  * localized for multiple languages :albania: :brazil: :bulgaria: :cn:
    :czech_republic: :de: :denmark: :es: :fr: :greece: :hungary: :india:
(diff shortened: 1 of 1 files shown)


goatshriek:
It is no trouble at all, thanks for submitting a fix.

I hadn't actually considered putting the code points directly into the README before, and I think this actually makes more sense than using the GitHub markdown name, which has confused myself and others before as well. I will probably move the README to do this at some point in the future.

For now to keep everything consistent, could you please update your change to use `:kr:` instead? This should render properly as 🇰🇷 .

goatshriek:
If you'd like, I can also make the change of all flags to the direct unicode characters as part of #425, in which case you can simply close this pull request. Either way forward is fine with me.

codehooni:
I think it would be nice if all the flags were changed to Unicode directly at once as part of #425 work, so I'll close that PR!

Issue #103, opened 2020-08-08
add support for colored terminal output


goatshriek:
A convenient feature in any logging library is colored output when logging to a console so that messages of different severity levels stand out from one another. This could be implemented in stumpless as a new set of attributes for stream loggers, which is off by default but is set when getting the stdout our stderr targets using `stumpless_open_stdout_target` or `stumpless_open_stderr_target` (respectively).

Bonus points if you can also implement build-time configuration options to set the default severity colors! However, if you aren't feeling up to this part then it can be split out into a separate issue for someone else to try their hand at - just make a note of this when you announce that you're going to work on this.

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with the stumpless code base and submitting changes.

## General Approach

There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here and someone can help you get past the stumbling block.

If you aren't familiar with printing colors in terminals, first do some research on the topic. [This article](https://solarianprogrammer.com/2019/04/08/c-programming-ansi-escape-codes-windows-macos-linux-terminals/) has a good walkthrough of the concepts for beginners.

`include/stumpless/target/stream.h` declares the public functions that are specific to stream targets. This is where you will need to add a new function to set the color for a severity code. Name the function something like `stumpless_set_severity_color` and will take a target, a severity code, and an escape code as arguments. Put the definition for this into the `src/target/stream.c` file.

Make sure that your implementation is thread safe! The simplest way to do this is to lock the target whenever it is being modified or read for the values, using the `lock_target` and `unlock_target` functions.

You'll need somewhere to store the configured colors. They should be added to the struct that stores details for stream targets, defined in `include/private/target/stream.h`. Exactly how the codes are stored is up to you, for example it could be an array of pointers to strings. Your new function for setting the color will update this field with the desired value. You should add code to `stumpless_open_stream_target` in `src/target/stream.c` to initialize these fields with something that means off (NULL pointers, perhaps)?

To set these up for stdout and stderr targets by default, just add some calls to your new set color functions to `stumpless_open_stderr_target` and `stumpless_open_stdout_target` (in `src/target/stream.c`) with some default colors that make sense. For example `STUMPLESS_SEVERITY_EMERG` should probably be something alarming like red!

[...]

PR #427, opened 2024-06-16
Terminal ansi colors

DR-Reg:
Adds support for using ansi escape codes to represent different severity levels when printing to a terminal, and presents some default escape codes. Allows for . Solves #103.

Unit tests were not written however, as I was unsure how to capture the stdout to check it was correctly printing ansi terminals (I did check with a script on my side that it was working).

This is my first open source contribution so any feedback is welcome.

### DIFF
--- a/include/stumpless/target/stream.h
+++ b/include/stumpless/target/stream.h
@@ -38,6 +38,7 @@
 #  include <stdio.h>
 #  include <stumpless/config.h>
 #  include <stumpless/target.h>
+#  include <stumpless/severity.h>
 
 #  ifdef __cplusplus
 extern "C" {
@@ -153,6 +154,32 @@ stumpless_open_stdout_target( const char *name );
 STUMPLESS_PUBLIC_FUNCTION
 struct stumpless_target *
 stumpless_open_stream_target( const char *name, FILE *stream );
+
+/**
+ * Sets the ANSI code to be printed in a specific target when a log is made at some severity
+ *
+ * It should be used with stdout or stderr as targets, since it is only for aesthetic purposes.
+ *
+ * **Thread Safety: MT-Safe race:name**
+ * This function is thread safe, of course assuming that escape_code is not modified by
+ * any other threads during execution.
+ *
+ * **Async Signal Safety: AS-Unsafe heap**
+ * This function is safe to call from signal handlers.
+ *
+ * **Async Cancel Safety: AC-Unsafe heap**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled.
+ *
+ * @param target The name of the target for which to set the colors.
+ *
+ * @param severity The severity code (LOG_ERR etc.) to which we set the specific color.
+ *
+ * @param escape_code The ANSI escape code representing the color.
+ */
+STUMPLESS_PUBLIC_FUNCTION
+void
+stumpless_set_severity_color( struct stumpless_target *target, enum stumpless_severity severity, const char *escape_code );
 
 #  ifdef __cplusplus
 }                               /* extern "C" */
--- a/include/stumpless/target/stream.h
+++ b/include/stumpless/target/stream.h
@@ -156,20 +156,22 @@ struct stumpless_target *
 stumpless_open_stream_target( const char *name, FILE *stream );
 
 /**
- * Sets the ANSI code to be printed in a specific target when a log is made at some severity
+ * Sets the ANSI escape code (https://en.wikipedia.org/wiki/ANSI_escape_code) to 
+ * be printed in a specific target when a log is made at some severity.
  *
  * It should be used with stdout or stderr as targets, since it is only for aesthetic purposes.
  *
- * **Thread Safety: MT-Safe race:name**
+ * **Thread Safety: MT-Safe race:escape_code**
  * This function is thread safe, of course assuming that escape_code is not modified by
  * any other threads during execution.
  *
- * **Async Signal Safety: AS-Unsafe heap**
- * This function is safe to call from signal handlers.
+ * **Async Signal Safety: AS-Unsafe lock**
+ * This function is not safe to call from signal handlers due to the destruction
+ * of a lock that may be in use.
  *
- * **Async Cancel Safety: AC-Unsafe heap**
- * This function is safe to call from threads that may be asynchronously
- * cancelled.
+ * **Async Cancel Safety: AC-Unsafe lock**
+ * This function is not safe to call from threads that may be asynchronously
+ * cancelled, as the cleanup of the lock may not be completed.
  *
  * @param target The name of the target for which to set the colors.
  *
--- a/src/target/stream.c
+++ b/src/target/stream.c
@@ -172,38 +172,35 @@ sendto_stream_target( struct stream_target *target,
   const char *reset_code = "\33[0m";
   unsigned short reset_code_len = strlen(sev_code);
   
+  config_lock_mutex( &target->stream_mutex );
   if (sev_code_len != 0)
   {
-    config_lock_mutex( &target->stream_mutex );
     fwrite_result = fwrite( sev_code, sizeof( char ), sev_code_len, target->stream );
-    config_unlock_mutex( &target->stream_mutex );
 
     if( fwrite_result != sev_code_len ) {
       goto write_failure;
     }
   }
 
-  config_lock_mutex( &target->stream_mutex );
   fwrite_result = fwrite( msg, sizeof( char ), msg_length, target->stream );
-  config_unlock_mutex( &target->stream_mutex );
   if( fwrite_result != msg_length ) {
     goto write_failure;
   }
 
   if (sev_code_len != 0)
   {
-    config_lock_mutex( &target->stream_mutex );
     fwrite_result = fwrite( reset_code, sizeof( char ), reset_code_len, target->stream );
-    config_unlock_mutex( &target->stream_mutex );
 
     if( fwrite_result != reset_code_len ) {
       goto write_failure;
     }
   }
 
+  config_unlock_mutex( &target->stream_mutex );
   return cap_size_t_to_int( fwrite_result + 1 );
 
 write_failure:
+  config_unlock_mutex( &target->stream_mutex );
   raise_stream_write_failure(  );
   return -1;
 }
--- a/src/target/stream.c
+++ b/src/target/stream.c
@@ -52,6 +52,8 @@ stumpless_close_stream_target( const struct stumpless_target *target ) {
 struct stumpless_target *
 stumpless_open_stderr_target( const char *name ) {
   struct stumpless_target *target = stumpless_open_stream_target( name, stderr );
+
+  if (target == NULL) return NULL;
   
   stumpless_set_severity_color(target, STUMPLESS_SEVERITY_EMERG_VALUE  , STUMPLESS_SEVERITY_EMERG_DEFAULT_COLOR  );
   stumpless_set_severity_color(target, STUMPLESS_SEVERITY_ALERT_VALUE  , STUMPLESS_SEVERITY_ALERT_DEFAULT_COLOR  );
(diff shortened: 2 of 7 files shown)


DR-Reg:
Do I need to 'sign the commit' so it can be merged? Or is anything else missing?

goatshriek:
Sorry for the delay in response, I have been focused on a different project for a bit.

First of all, thank you for tackling this issue! As I am sure you saw in the issue history it has drawn many contenders but remains yet unconquered.

Signing commits is not a requirement for PRs here, particularly since they are typically squashed anyway. However, I do recommend you work through that process for your own education (if education is something you're into, at least). But don't worry about signing and re-pushing your work or anything like that.

At first glance this seems like a solid contribution that shouldn't need much re-shaping, but I need to set aside time to sit down and review it in detail. I will get back to you within the next week with a review.

include/stumpless/target/stream.h
```diff
@@ -154,6 +155,32 @@ STUMPLESS_PUBLIC_FUNCTION
 struct stumpless_target *
 stumpless_open_stream_target( const char *name, FILE *stream );
 
+/**
+ * Sets the ANSI code to be printed in a specific target when a log is made at some severity
+ *
+ * It should be used with stdout or stderr as targets, since it is only for aesthetic purposes.
+ *
+ * **Thread Safety: MT-Safe race:name**
```
goatshriek: update the race variable name to `escape_code`

include/stumpless/target/stream.h
```diff
@@ -154,6 +155,32 @@ STUMPLESS_PUBLIC_FUNCTION
 struct stumpless_target *
 stumpless_open_stream_target( const char *name, FILE *stream );
 
+/**
```
goatshriek, line 158: This function documentation is a good place to include a link to information on ANSI escape codes, perhaps the Wikipedia article for "ANSI escape code"?

src/target/stream.c
```diff
@@ -85,6 +118,25 @@ stumpless_open_stream_target( const char *name, FILE *stream ) {
   return NULL;
 }
 
+void
+stumpless_set_severity_color( struct stumpless_target *target, enum stumpless_severity severity, const char *escape_code )
+{
+  if (severity > 7)
```
goatshriek: use `severity_is_invalid` here instead

src/target/stream.c
```diff
@@ -85,6 +118,25 @@ stumpless_open_stream_target( const char *name, FILE *stream ) {
...
+    raise_index_out_of_bounds("Severity value must be part of 'stumpless_severity' enum (0-7)", severity);
```
goatshriek: similarly, use `raise_invalid_severity` here instead

src/target/stream.c
```diff
@@ -85,6 +118,25 @@ stumpless_open_stream_target( const char *name, FILE *stream ) {
...
+
+  if (target->type != STUMPLESS_STREAM_TARGET)
+    raise_target_unsupported("This function is only supported for stream targets");
+
+  struct stream_target *starget = (struct stream_target *)target->id;
+  lock_target(target);
```
goatshriek: acquire this lock before checking the type, so that it is locked for the duration of the code that uses it

include/stumpless/target/stream.h
```diff
@@ -154,6 +155,32 @@ STUMPLESS_PUBLIC_FUNCTION
...
+ * Sets the ANSI code to be printed in a specific target when a log is made at some severity
+ *
+ * It should be used with stdout or stderr as targets, since it is only for aesthetic purposes.
+ *
+ * **Thread Safety: MT-Safe race:name**
+ * This function is thread safe, of course assuming that escape_code is not modified by
+ * any other threads during execution.
+ *
+ * **Async Signal Safety: AS-Unsafe heap**
```
goatshriek: this is both `AS-Unsafe` and `AC-Unsafe` due to lock usage, but I don't see any heap usage. Be sure you update the verbiage below the designations as well.

src/target/stream.c
```diff
@@ -111,17 +163,44 @@ new_stream_target( FILE *stream ) {
...
   size_t fwrite_result;
+  enum stumpless_severity severity = stumpless_get_entry_severity(entry);
+  const char *sev_code = target->escape_codes[severity];
+  unsigned short sev_code_len = strlen(sev_code);
+  const char *reset_code = "\33[0m";
+  unsigned short reset_code_len = strlen(sev_code);
+  
+  if (sev_code_len != 0)
+  {
+    config_lock_mutex( &target->stream_mutex );
```
goatshriek: you'll probably need to expand this from three individual critical sections to a single larger one, locking before the first write and only unlocking after the reset code is written. Otherwise, writes may end up being interleaved between messages with strange resulting output.

src/target/stream.c
```diff
@@ -111,17 +163,44 @@ new_stream_target( FILE *stream ) {
...
                       const char *msg,
-                      size_t msg_length ) {
+                      size_t msg_length,
+                      const struct stumpless_entry *entry ) {
   size_t fwrite_result;
+  enum stumpless_severity severity = stumpless_get_entry_severity(entry);
+  const char *sev_code = target->escape_codes[severity];
+  unsigned short sev_code_len = strlen(sev_code);
+  const char *reset_code = "\33[0m";
+  unsigned short reset_code_len = strlen(sev_code);
```
goatshriek: should the argument to `strlen` be `reset_code` here instead?

src/target/stream.c
```diff
@@ -85,6 +118,25 @@ stumpless_open_stream_target( const char *name, FILE *stream ) {
...
+    raise_target_unsupported("This function is only supported for stream targets");
+
+  struct stream_target *starget = (struct stream_target *)target->id;
+  lock_target(target);
+
+  strncpy(starget->escape_codes[severity], escape_code, 32);
+  starget->escape_codes[severity][31] = 0;
+
+  unlock_target(target);
+}
```
goatshriek, line 143: be sure you call `clear_error` before returning so that the error code is set to clean in the event of success

src/target/stream.c
```diff
@@ -85,6 +118,25 @@ stumpless_open_stream_target( const char *name, FILE *stream ) {
...
 }
 
+void
+stumpless_set_severity_color( struct stumpless_target *target, enum stumpless_severity severity, const char *escape_code )
+{
+  if (severity > 7)
+    raise_index_out_of_bounds("Severity value must be part of 'stumpless_severity' enum (0-7)", severity);
+
+  if (target->type != STUMPLESS_STREAM_TARGET)
+    raise_target_unsupported("This function is only supported for stream targets");
```
goatshriek, line 132: This error message should be a localized string. The documentation for [localization](https://github.com/goatshriek/stumpless/blob/latest/docs/localization.md#adding-new-strings) describes how to add new localized strings: you just need to run `scripts/add_l10n_string.rb` with your new string.

goatshriek (changes requested):
Thanks again for putting this together, and your patience in waiting for a review. This is great work! The changes below should be relatively straightforward.

However, you will also need to write some tests for this functionality. This will be a little more involved - the way I would approach this is to create a file, and use the stream to create a new stream target that writes into it. You can then check the file contents after the writing to make sure that the expected escape codes are there (or not, depending on the test).

include/stumpless/target/stream.h
DR-Reg: Done.

include/stumpless/target/stream.h
DR-Reg: Done.

src/target/stream.c
DR-Reg: Done.

src/target/stream.c
DR-Reg: Done.

src/target/stream.c
DR-Reg: Done.

include/stumpless/target/stream.h
DR-Reg: Done.

src/target/stream.c
DR-Reg: Done.

src/target/stream.c
DR-Reg: Yes. Done.

src/target/stream.c
DR-Reg: Done.

src/target/stream.c
DR-Reg: Thank you for all the feedback! I have been unable to add the localised string due to what may be a bug in the script (output shown below). I am not very fluent in ruby so I decided to just leave it as is for somebody else to work on.

Output:
```
$ ruby scripts/add_l10n_string.rb SET_SEVERITY_COLOR_UNSUPPORTED_TARGET_ERROR_MESSAGE This function is only supported for stream targets
scripts/add_l10n_string.rb:59:in `each': Is a directory @ io_fillbuf - fd:7 <home-path>/stumpless/include/private/config/locale/. (Errno::EISDIR)
        from scripts/add_l10n_string.rb:59:in `block in <main>'
        from scripts/add_l10n_string.rb:53:in `each'
        from scripts/add_l10n_string.rb:53:in `<main>'
```

Edit: forgot to mention it is ruby 3.3

goatshriek: Ah, that is unfortunate. Thank you very much for sharing the details though, they did allow me to identify the issue and a fix. You can leave this out and I will go back and correct it afterwards - line 53 of the script needs to be changed to `each_child`. I'll submit a separate patch later.

goatshriek:
You're closing in on done! Just two more changes needed - you'll need to resolve the Windows test failures, which will probably require some refactoring of the tests themselves to be a bit more portable. You may need to create a new file instead of using the stderr/stdout targets, but I'll leave this up to you.

Also, please add a test where you attempt to call `stumpless_set_severity_color` with an incompatible target type, to make sure that this fails as expected.

DR-Reg:
I have added some preprocessor directives so the syscalls to dup etc. are conformant with the required windows ISO C rules. In the end I thought it was better to redirect stdout/stderr since in theory (as I understand) files should not have these escape codes (unless set manually).

I am aware that it may be possible to make this code simpler using some sort of macro to distinguish between code needed for windows vs unix. (however I wasn't sure what your policy on macros is, I can rewrite the code in this way if you want).

Also if you think the file approach is cleaner I will rewrite the test.

Also note that for some reason running on my windows machine I get that the regex expression is invalid? Is this a bug in the windows version of gtest or my fault?

Sorry for the inconvenience.

Edit: PS. Should the test for `stumpless_set_severity_color` be in test/function/target/stream.cpp or elsewhere?

goatshriek:
The tests have a more relaxed standard for macros and wrapping up OS-specific functionality, so I think what you've done here is fine, no need to rewrite at this point. As long as they run on the major tested systems (which are currently the ones covered by CI) then I think we're good to go.

Hmm, I'm not sure why the Windows regexes are not working properly - perhaps the underlying implementation is more strict about regular expressions with non-displayable characters in them? Does changing the escape character to `\x1b` instead of the octal value help?

Yes, I think that `test/function/target/stream.cpp` is the best place for the set severity color tests.

DR-Reg:
In the end the regex bug was due to gtest having limited regex support on windows (as seen [here](https://google.github.io/googletest/advanced.html#regular-expression-syntax)). I assume they will add better support in the future, so I put in a bit of a 'janky' solution for windows using one of GTEST's processor macros (so when they do add the functionality the actual regex is called).
Is this ok?

goatshriek:
This will be okay, though I will ask that you please add a comment to the test code before the preprocessor conditional that explains why this is done so that others reading the tests will know why this is being done without finding this conversation.

It looks like you just need to update the `src/windows/stumpless.def` file with the new function, and address some memory leaks that valgrind has identified, and you'll be closing in on done.

goatshriek (approved):

goatshriek:
Thank you again for working through this, and I hope that you feel a sense of accomplishment for completing an issue that gave so many others quite a bit of trouble!

I try to give contributors a shoutout on X when I can, especially for larger contributions like this one. I don't see one listed on your profile here though: if you have an account let me know (via email or private Gitter if you want it to remain hidden) and I will post something there.

PR #432, opened 2024-07-29
add benchmark for stumpless_get_prival_string

### LINKED ISSUE #431: add benchmark for stumpless_get_prival_string
Some future changes will focus on improving the performance of the to string functions, including this one. Adding a benchmark test for `stumpless_get_prival_string` will give future changes a way to evaluate what the impact to performance is.

## General Approach

First, read the [documentation](https://github.com/goatshriek/stumpless/blob/latest/docs/benchmark.md) on benchmarking in stumpless. This doc walks through the general philosophy behind benchmark tests and how they are used in this project.

Once you understand these concepts, you'll be ready to add a new benchmark in `test/performance/prival.cpp` for `stumpless_get_prival_string`. Look to the existing test suites as a starting point for your new one, and make sure that you are able to see results from your new test when you run the prival benchmark suite (the `run-performance-test-prival` build target does this for you).

amitansu1024:
Adds benchmarking for the function stumpless_get_prival_string. fixes #431

### DIFF
--- a/test/performance/prival.cpp
+++ b/test/performance/prival.cpp
@@ -22,6 +22,7 @@
 #include "test/helper/memory_counter.hpp"
 
 NEW_MEMORY_COUNTER( prival_from_string )
+NEW_MEMORY_COUNTER( get_prival_string )
 
 static void PrivalFromString(benchmark::State& state) {
   // Create a list of priorities.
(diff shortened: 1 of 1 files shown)


test/performance/prival.cpp
```diff
@@ -44,4 +45,25 @@ static void PrivalFromString(benchmark::State& state) {
...
+
+  const char *result;
+  INIT_MEMORY_COUNTER( get_prival_string );
+
+  for (auto _ : state) {
+    for (auto prival: prival_list ) {
+      result = stumpless_get_prival_string( prival );
+    }
+  }
+  free( ( void * ) result );
```
goatshriek: This call to `free` will need to move to inside both loops to avoid a memory leak in this test.

goatshriek (changes requested):
Thanks for putting this together! I believe two changes will be enough for this to be accepted: moving `free` as noted below, and adding the header for `free` as well.

I'm unsure why the locale CI builds are failing but not others. Let's see if the other changes resolve this failure as well.

goatshriek (approved):

goatshriek:
Thanks again! I've identified the issue affecting some of the builds - this will get merged once that makes it in so that this change doesn't falsely appear to cause failures.

Issue #428, opened 2024-06-30
refactor prival from string for performance


goatshriek:
While the `stumpless_prival_from_string` function works, it currently allocates and later frees heap memory for temporary storage of strings during parsing. Now that #385 and #383 have been implemented, this function should be refactored so that it uses the new `_from_buffer` functions for facility and enum parsing to avoid heap memory allocations.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

Review the documentation for `stumpless_get_facility_enum_from_buffer` and `stumpless_get_severity_enum_from_buffer` so that you understand how they work, and then understand how `stumpless_prival_from_string` in `src/prival.c` currently works. This is the function you will be reworking.

Before you make any changes, make sure that you can run the performance tests, specifically the `PrivalFromString` benchmark. Make sure that you can collect performance both _before_ and _after_ you implement your change. You should see CPU cycles and memory allocated decrease as a result - if you do not, review your implementation and find out why. You can find details on running the performance benchmarks in stumpless in the [docs/benchmark.md](https://github.com/goatshriek/stumpless/blob/latest/docs/benchmark.md) documentation.

Once you have this, make your changes and make sure that performance improves as expected! The test suites should continue to pass and coverage should not decrease as a result of your changes - you may need to add more tests for the refactored function to `test/function/prival.cpp` improve coverage depending on your implementation.

Griezn:
Hey, I would like to solve this issue. Could you assign me to it?

goatshriek:
Sure thing, good luck!

PR #437, opened 2024-08-17
Refactored prival from string for performance

Griezn:
This PR addresses issue #428 by significantly reducing allocations and execution time, as demonstrated by the `run-performance-test-prival` results.

Updated `get_x_from_buffer` functions to correctly utilize the length parameter and ensure function test correctness.

The check I removed failed. I don't know exactly why it the check didn't exist in the corresponding test of facility so I removed it.

### DIFF
--- a/src/prival.c
+++ b/src/prival.c
@@ -24,7 +24,6 @@
 #include <stumpless/prival.h>
 #include <stumpless/severity.h>
 #include <stumpless/facility.h>
-#include <stumpless/error.h>
 #include "private/config.h"
 #include "private/config/wrapper/locale.h"
 #include "private/facility.h"
@@ -32,6 +31,7 @@
 #include "private/prival.h"
 #include "private/severity.h"
 #include "private/validate.h"
+#include "private/error.h"
 
 const char *
 stumpless_get_prival_string( int prival ) {
--- a/src/prival.c
+++ b/src/prival.c
@@ -114,7 +114,7 @@ stumpless_prival_from_string( const char *string ) {
 
   severity = stumpless_get_severity_enum_from_buffer( period + 1, len );
 
-  if( severity < 0 ) {
+  if( severity < 0 && !stumpless_has_error()) {
     raise_invalid_param(  );
     return -1;
   }
--- a/src/prival.c
+++ b/src/prival.c
@@ -64,7 +64,6 @@ stumpless_prival_from_string( const char *string ) {
   int prival;
   int severity;
   int facility;
-  const char *param;
   const char *period;
   const char *sec_period;
   size_t len;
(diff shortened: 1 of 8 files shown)


test/function/prival.cpp
```diff
@@ -134,8 +134,6 @@ namespace {
     result = stumpless_prival_from_string( "syslog.err" );
     EXPECT_EQ( result, -1 );
 
-    EXPECT_ERROR_ID_EQ( STUMPLESS_MEMORY_ALLOCATION_FAILURE );
```
goatshriek, line 137: If this line causes the test to fail, it seems like the memory allocation failure error is getting overwritten before the return. Since this would be misleading to the caller, please find a way to preserve the error instead. I'm guessing the check and error raising at lines 117 and 118 of `src/prival.c` are the cause, but haven't confirmed by testing it.

goatshriek (changes requested):
Thanks very much for putting this change together! My apologies for the delay in getting it reviewed.

There is just one change that needs to be made in order to properly preserve the cause of an error, and once that is done this should be ready to go!

goatshriek:
Just one more header fixup and I believe this will be ready.

goatshriek (dismissed):

src/prival.c
```diff
@@ -122,17 +113,9 @@ stumpless_prival_from_string( const char *string ) {
...
-    return -1;
-  }
-
-  severity = stumpless_get_severity_enum( param );
-
-  free_mem( param );
+  severity = stumpless_get_severity_enum_from_buffer( period + 1, len );
 
-  if( severity < 0 ) {
+  if( severity < 0 && !stumpless_has_error()) {
```
goatshriek: rather than an extra part of this conditional, please separate the "has_error" check so that it is in a separate conditional that simply returns (without raising a new error) if an error is present. As is, this branch is skipped in the memory allocation failure case, and the response is only -1 as a side effect of `get_prival` being the result of a bitwise or with -1 rather than being a true error.

goatshriek (changes requested):
My apologies for changing this review response, but in reviewing the coverage report and seeing the error check as only partially covered, I believe the logic needs to be adjusted slightly as mentioned in the comment. This should both correct the coverage and also make the error return path more clear and intentional.

Griezn:
So what I did now is, I solved issue #438 also with this pr. With this fix there are no allocations performed by the function so from my perspective those tests become irrelevant. This also allowed me to remove the check I added to see if there already was an error.

But I think the issue of errors being overwritten can still occur in other places so maybe this is a new issue.

goatshriek (changes requested):
I think this is a great approach! There are a few other tests that are failing now as they were expecting memory allocation failures - you can either remove these completely or simply change the expected results to be no error.

goatshriek:
The Windows builds are failing because `strncasecmp` isn't completely portable, it seems. I can think of two ways forward right now. The first is a bit simpler: to add a custom implementation of `strncasecmp` to the `strhelper` files (probably with a different function name) and use this. The second is more involved but will be the end state of the project: add a configuration step to detect if the build environment has a library function that can be used (`strncasecmp`, `_strnicmp`, etc) and uses this if it is available, and otherwise falls back to an embedded default.

Since the second option will require some cmake work and the addition of several more source files and `config_` functions (see the [portability docs](https://github.com/goatshriek/stumpless/blob/latest/docs/portability.md) for more info), I would just stick with the first option in this change. I can create a second issue to track the second part of this, since using the system library would reduce library size and potentially improve performance when available.

goatshriek (approved):

goatshriek:
Thanks once again for putting this change together, and sticking with it through all of the requested changes!

Issue #444, opened 2024-09-26
update benchmark documentation to match code


goatshriek:
The documentation for benchmarking includes a [walkthrough](https://github.com/goatshriek/stumpless/blob/latest/docs/benchmark.md#walkthrough-improving-stumpless_copy_element) of running performance tests on a piece of code, refactoring it, and testing for improvement.

While the content of the walkthrough is still relevant, the code snippet included for the performance test itself has not been updated to match the current performance test template, which is simpler to read and understand.

This is a simple change: update the benchmarking documentation code snippet for the `CopyElement` test to match the implementation in `test/performance/element.cpp`.

If you are up for it, refactor the test to make it even simpler by using the fixture function `create_entry` to construct an entry with elements, rather than doing this manually. This is optional though, and if you would rather focus on the mechanics of making changes to files and issuing a pull request, you are free to leave this change out.

abdalahsalah:
@goatshriek need review.

vididvidid:
Hi 👋,
I’d love to contribute to this issue as a first-time contributor. Could you please assign this to me or open it up for a PR? I’d be happy to work on the implementation following the discussion above.

Thanks! 🙏

goatshriek:
Sure thing. I also recommend you go through the previously submitted PRs for this issue so that you can see what caused others to get stuck, so that you can avoid repeating their mistakes.

prasadsarwar:
Hello! I have my development environment set up and I would like to work on this issue as my first contribution. Could you please assign it to me?

goatshriek:
Hi @prasadsarwar, feel free to give this one a shot. Given how many unsuccessful/ghosted attempts on this one there have been, I will not be assigning it at this point until there is an actual PR submitted.

prasadsarwar:
Hi, thanks for the clarification. I will start working on this now and will aim to submit a PR. Thanks for the opportunity!

goatshriek:
Resolved via #549.

PR #452, opened 2024-09-28
update benchmark documentation to match code

Brijeshthummar02:
issue #444

### DIFF
--- a/docs/benchmark.md
+++ b/docs/benchmark.md
@@ -72,14 +72,12 @@ static void CopyElement(benchmark::State& state){
   stumpless_set_realloc( copy_element_memory_counter_realloc );
   stumpless_set_free( copy_element_memory_counter_free );
 
-  element = stumpless_new_element( "copy-element-perf" );
-  stumpless_add_new_param( element, "param-1", "value-1" );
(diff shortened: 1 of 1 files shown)


docs/benchmark.md
```diff
@@ -72,14 +72,12 @@ static void CopyElement(benchmark::State& state){
   stumpless_set_realloc( copy_element_memory_counter_realloc );
   stumpless_set_free( copy_element_memory_counter_free );
 
-  element = stumpless_new_element( "copy-element-perf" );
-  stumpless_add_new_param( element, "param-1", "value-1" );
-  stumpless_add_new_param( element, "param-2", "value-2" );
+  element = create_entry("copy-element-perf");
```
goatshriek, line 75: Please review the signature for `create_entry`. It does not take an argument.

goatshriek (changes requested):
Please review the actual test in `test/function/element.cpp`, this change does not completely bring the two into matching.

Thank you for refactoring with the fixture function! Be sure that you update the actual test as well as the docs with the refactored code, to ensure that it compiles and works correctly.

Issue #445, opened 2024-09-26
fix custon strncasecmp implementation


goatshriek:
Stumpless currently includes an implementation of `strncasecmp` called `strncasecmp_custom` in `src/strhelper.c`. This function is used to perform case-insensitive comparison of strings in order to derive stumpless types from strings to support things like CLI development.

However, this implementation currently has a flaw where a string that is a prefix of another will be incorrectly returned as equal. You will add a failing test that catches this behavior, and correct it.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

Adding a failing test to `test/function/severity.cpp` for this case is straightforward. Create a new test based on `TEST( GetSeverityEnum, NoSuchSeverity )` named something like "IncompleteSeverity" that passes a prefix of a valid severity, such as "not". This should return -1, but currently will instead return `STUMPLESS_SEVERITY_NOTICE`.

Once you have added this test and ensured that it fails, correct the implementation of `strncasecmp_custom` so that this no longer occurs. Verify that your new test now passes, as well as all of the existing tests as well.

goatshriek:
Fixed by #454.

PR #454, opened 2024-09-29
Issue #445. Testing the behavior of the strncasecmp_custom function. 

lecapitainenemo:
Related to issue #445 fix custom strncasecmp implementation.

It was described that the strncasecmp_custom function was going to fail and emit that strings like "not" and "NOTICE" were equal. However, the current implementation passed tests of this style (I added them).

This was a not a notable change so I did not add anything to ChangeLog.

Thanks.

### DIFF
--- a/src/strhelper.c
+++ b/src/strhelper.c
@@ -89,7 +89,7 @@ strncasecmp_custom( const char *s1, const char *s2, size_t n ) {
       if (*s1++ == '\0')
         break;
     } while (--n != 0);
-    if(*s2 != '\0') return tolower(*s1) - tolower(*s2);
+    if(*s2 != '\0') return tolower(*s2) - tolower(*s1);
   }
   return 0;
 }
--- a/src/strhelper.c
+++ b/src/strhelper.c
@@ -85,10 +85,11 @@ strncasecmp_custom( const char *s1, const char *s2, size_t n ) {
   if (n != 0) {
     do {
       if (tolower(*s1) != tolower(*s2++))
-        return tolower(*s1) - tolower(*--s2);
+        return tolower(*--s2) - tolower(*s1);
       if (*s1++ == '\0')
         break;
     } while (--n != 0);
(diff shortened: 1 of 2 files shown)


lecapitainenemo:
I was wondering how to proceed, if I committed a mistake or if there is any other recommended task.
Thank you

goatshriek:
The bug can be subtle, but is present. By passing in a buffer size far greater than the string you provided, you've forced the comparison to go farther in and find the NULL character difference. However the from_buffer function needs to work when it doesn't have a NULL terminator. After all the purpose of this function is to NOT need a NULL terminator to begin with!

You can trigger the failure by reducing the buffer size to match the string length, add this test, or [my recommendation] do both.

```cpp
TEST( GetSeverityEnum, InvalidButPrefix ) {
  enum stumpless_severity result = stumpless_get_severity_enum( "war" );
  EXPECT_EQ( result, -1 );
  EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_SEVERITY );
}
```

lecapitainenemo:
Hey,
I added the failing tests. I removed the "EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_SEVERITY );" part from them, though. With the modification to strncasecmp_custom function, all the tests (new and old) pass. Please let me know if there is anything wrong. I appreciate your time and patience.

goatshriek:
Ah, I didn't realize that error codes were not being raised. I've opened #455 to address that.

Everything looks good! One final ask: can you also add a new test for the opposite direction: that strings that start with a valid severity but are longer are not accepted. For example "warnings are neat" should not be parsed to a valid severity. This will make sure that the comparison works properly in both directions.

lecapitainenemo:
The respective tests were added and all old and new passed.
Thank you

goatshriek:
Thanks very much for putting this contribution together and sticking with it through the requested changes!

Issue #440, opened 2024-09-26
add alloc_array function


goatshriek:
Stumpless includes wrappers for memory allocation to allow a customized memory allocator to be used. The primary function used for this is the `alloc_mem` function, which calls the underlying allocator, which is `malloc` by default.

A frequent idiom in the code base is to use `alloc_mem` with the result of a multiplication of a count of items by their size. Consider this line taken from `src/entry.c` in the `stumpless_copy_entry` function:

```c
copy->elements = alloc_mem( entry->element_count * sizeof( element_copy ) );
```

There are two things here that can be improved. First and most importantly, there is not overflow protection on the multiplication. If `element_count` causes overflow when multiplied with the size of the element copy, this memory block could be too small, and will almost certainly result in heap corruption. Second, this pattern is used in other places in the code base, and adding overflow checks to each one leaves the potential to miss one. It would be better to consolidate this logic.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, read [this](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions) section of the development documentation on adding new functions. Once you understand how this process works, start defining the new function.

Add a new function to `src/memory.c` named `alloc_array` with the following signature:

```c
void *
alloc_array( size_t item_count, size_t item_size );
```

This function will perform a multiplication overflow check, and if this passes call `alloc_mem` with the correct size. If it fails, it will raise a memory allocation failure just as `alloc_mem` does and return NULL.

If you need an example of overflow detection, have a look at [musl's implementation of `calloc`](https://git.musl-libc.org/cgit/musl/tree/src/malloc/calloc.c), which includes the same check. While `alloc_array` does not need to zero memory, the overflow detection will be the same.

Once you've made this change, swap out the usage in `src/entry.c` mentioned above, and make sure that all tests continue to pass.

Finally, add a new check to ensure that the overflow detection actually works. This should go in the test suite for the function that you updated to use the new function, which is in `test/function/entry.cpp`. The simplest way to achieve this is to create an entry, and manually update the `element_count` field to be something sufficiently large, like `SIZE_MAX`.

aadit-n3rdy:
Hi, can I be assigned this issue?

goatshriek:
Closed by #458.

PR #458, opened 2024-10-05
Adding an `alloc_array` function

aadit-n3rdy:
This PR solves #440 and adds an `alloc_array` function. All uses of `alloc_mem` which can be replaced with `alloc_array` have been replaced with `alloc_array`. A test for `stumpless_copy_entry` has been added as well.

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -130,8 +130,8 @@ realloc_mem( const void *mem, size_t size ) {
 void *
 alloc_array( size_t item_count, size_t item_size ) {
   if (item_count && item_count >= (size_t)-1/item_size) {
-	raise_memory_allocation_failure();
-	return NULL;
+    raise_memory_allocation_failure();
+    return NULL;
   }
   return alloc_mem(item_count * item_size);
 }
--- a/src/memory.c
+++ b/src/memory.c
@@ -126,3 +126,12 @@ realloc_mem( const void *mem, size_t size ) {
 
   return new_mem;
 }
+
+void *
+alloc_array( size_t item_count, size_t item_size ) {
+  if (item_count && item_count >= (size_t)-1/item_size) {
+	raise_memory_allocation_failure();
+	return NULL;
+  }
+  return alloc_mem(item_count * item_size);
+}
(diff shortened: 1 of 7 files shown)


aadit-n3rdy:
Both the failed tests were in calls to `alloc_mem`, is this a problem with `alloc_array`? I'll look into this as well.

EDIT: Scratch that, forgot to free the entry I created in the test. My bad.

src/memory.c
```diff
@@ -126,3 +126,12 @@ realloc_mem( const void *mem, size_t size ) {
 
   return new_mem;
 }
+
+void *
+alloc_array( size_t item_count, size_t item_size ) {
+  if (item_count && item_count >= (size_t)-1/item_size) {
+	raise_memory_allocation_failure();
+	return NULL;
```
goatshriek: it looks like there are some tabs here; please convert them to spaces

test/function/entry.cpp
```diff
@@ -359,6 +371,14 @@ namespace {
     EXPECT_TRUE( set_malloc_result == malloc );
   }
 
+  TEST_F( EntryTest, CopyMallocFailureOnSizeOverflow ) {
+	  const struct stumpless_entry *result;
+
+	  result = stumpless_copy_entry( large_entry );
+	  EXPECT_NULL( result );
+	  EXPECT_ERROR_ID_EQ( STUMPLESS_MEMORY_ALLOCATION_FAILURE );
```
goatshriek: looks like there are some tabs here as well

test/function/entry.cpp
goatshriek, line 1: please move the creation/deletion of the `large_struct` out of the test fixture and into the single new test you've made, changing it from `TEST_F` to `TEST` in the process.

The structures in the fixture are intended to be valid, and since this one is a bit strange we only need to create it for this one specialized test.

goatshriek (changes requested):
Excellent work! Just a few cleanup items: make sure that you aren't using tabs anywhere, and move the large struct into the new test instead of making it a fixture structure.

test/function/entry.cpp
aadit-n3rdy: Done, I've moved it to a separate test named `CopyEntry, MallocFailureOnSizeOverflow`. Does that work?

goatshriek (approved):

PR #460, opened 2024-10-06
Updating the pull request

Sanidhyafeaturist:


goatshriek:
There are multiple problems here. This code obviously won't compile, making me think you didn't even make an attempt to run this. `stumpless_buffer_t` doesn't even exist. The new benchmark is at a different path than the issue specifies. You clearly did not look at the existing benchmark tests, and make no use of the benchmark framework they use, instead using a hand-rolled timing-based scheme.

Quite simply, it looks like you threw this issue into an LLM and pasted the output into your PR without so much as compiling it. Respect maintainers' time and only submit quality code that you have tested and believe is ready for consideration, and read [this language](https://github.com/goatshriek/stumpless/blob/typo-fix/docs/CONTRIBUTING.md#a-note-on-ai) on the use of AI tools in this project.

This is your second submission of low-quality code within 30 minutes even after a warning to submit when ready for review. A third instance of this behavior will result in a temporary ban as per the [Code of Conduct](https://github.com/goatshriek/stumpless/blob/latest/docs/CODE_OF_CONDUCT.md).

Issue #117, opened 2020-09-12
add a new (spoken) language


goatshriek:
If you know a language other than English and you're looking for a way to contribute to a project without knowing much about coding, adding a new locale is a great place to start.

There is a detailed guide on how to go about adding a new language that you know in the [localization documentation](https://github.com/goatshriek/stumpless/blob/latest/docs/localization.md). If you're feeling up to it, go ahead and open a new issue with the language you would like to add, and go to it!

Is the language that you wanted already added? Look over the locale header for it anyway (in `include/private/config/locale/`); any strings added after it was initially added will still be English placeholders marked with a `// todo translate` comment. A contribution translating these is almost as important as the initial translation, and is a much simpler change. Rather than implementing a completely new set of headers, you will just change the placeholder strings to be locale-specific strings instead. If you want to see a list of what strings need translating by locale, you can also use grep to get a current list:

```sh
grep -r -c "// todo translate" include/private/config/locale/

# output looks like this:
# include/private/config/locale/bg-bg.h:28
# include/private/config/locale/cz-cz.h:15
# include/private/config/locale/de-de.h:36
# include/private/config/locale/el-gr.h:16
# include/private/config/locale/en-us.h:0
# include/private/config/locale/es-es.h:18
# include/private/config/locale/fr-fr.h:36
# include/private/config/locale/it-it.h:0
# include/private/config/locale/pl-pl.h:17
# include/private/config/locale/sk-sk.h:31
# include/private/config/locale/sv-se.h:36
```

If there aren't any for your language or you'd rather do something new, have a look in the `l10n` folder of the repository. This has localized documentation for the project, and there are lots of opportunities to add a paragraph or two of translation. Don't feel like you need to translate an entire file for this, just a few paragraphs is more than enough! The `l10n/zh-cn` folder is a good example of what this could look like as more translations are added.

Lockna:
I would like to add a German translation ^^

goatshriek:
@Lockna sounds good! I have created #121 for the specific locale. I also added a reference to a recent commit that has examples of a similar update recently, if you want a clean example to complete the documentation. Any suggestions on improvements to the documentation/process are also welcome!

pepo939393:
I would like to add Slovak translation

goatshriek:
@pepo939393 Awesome! I just created #137 for you, it has a reference to the most recent locale addition as an example.

taliezincho:
I'll try to add Bulgarian :)

best regards

[...]

PR #462, opened 2024-10-06
Translation of all german messages (#117)

Testspieler09:
Translated all english messages marked as `// todo translation`
Referring to issue: #117


goatshriek (changes requested):
Thanks very much for updating this translation - there have been quite a few new strings since it was first done.

Only a few lines that need to be trimmed to be less than 80 characters: they are noted in the "Static Analysis" CI run. Once those are fixed up this should be good to go!

Testspieler09:
Thank you for the feedback I shortened the lines now (a3e9c0833c69139578520979228d8bfa94d54ea1). If there is anything else I can do to improve this PR let me know.

Issue #442, opened 2024-09-26
test cache page allocation failure


goatshriek:
Stumpless uses a simple slab cache implementation in `src/cache.c` to consolidate memory allocations for common structures like entries. [Test coverage](https://app.codecov.io/gh/goatshriek/stumpless/blob/latest/src%2Fcache.c) of this cache is fairly complete, but still has some gaps.

One such gap is the failure of a new page allocation when the cache needs one. You will test for this case by causing a memory allocation failure after the cache has been set up and needs to add a new page.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, familiarize yourself with the `cache_alloc` implementation in `src/cache.c`. This function looks for a free slot in the cache, and if one is not found then it allocates a new page to make more space. You don't need to fully understand this function, but it will help to see what logic causes the `add_page` call to happen, which is what you will cause to fail.

Next, devise a test that creates a single structure in a cache (entry structures are one such type) to set up the cache, set the memory allocation function to one that will fail, and then create new structures until the cache fills, attempts to allocate a new page, and fails. For an example of a test that creates several new structures look at `TEST( BufferTargetOpenTest, Open100Targets )` in `test/function/target/buffer.cpp`, and `TEST_F( EntryTest, AddNewParamMallocFailure )` in `test/function/entry.cpp` for an example of a memory allocation failure test. If testing entry creation, put your new test in `test/function/entry.cpp`.

Make sure that you have full coverage on `add_page` in `src/cache.c` after this test. See the [test documentation](https://github.com/goatshriek/stumpless/blob/latest/docs/test.md#test-coverage) for details on checking coverage.

rrrrrrmb:
Hi, I'd like to work on this. I haven't really done much in C but would like to try. Based on my understanding  the line of interest to us is 

`new_page = alloc_mem( c->page_size );` in `add_page` in `src/cache.c`

This new page allocation should fail in our case however I'm not sure how to set this up with `MALLOC_FAIL_ON_SIZE` since I don't know how to get the page_size value. Is there another way to do this ? Can you give pointers to solve this ? 

Thanks !

goatshriek:
It is a tricky one to understand and test for, which is why there isn't a test now!

[...]

PR #464, opened 2024-10-07
test cache page allocation failure

rrrrrrmb:
This PR solves #442. This test creates an `entry` structure to initialize the cache and then creates more structures to  fill the cache until a new page allocation fails.

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -71,6 +71,7 @@ add_page( struct cache *c ) {
   if( !new_pages ) {
     return -1;
   }
+  c->pages = new_pages;
 
   new_page = alloc_mem( c->page_size );
   if( !new_page ) {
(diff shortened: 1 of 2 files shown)


goatshriek (changes requested):
Your test works perfectly, and has in fact uncovered a memory leak!

The result of `realloc_mem` on line 70 of `src/cache.c` isn't dealt with properly in the case of failure. The easiest solution here would be to move the assignment currently on line 83 to before the alloc call, so that the cache has the new structure even in the case of an error. If you'd like to give that a shot please do, or if you see a better solution you can try that instead.

rrrrrrmb:
Thanks for the hint ! I've made the suggested change and looks like that fixed it :)

goatshriek (approved):

Issue #449, opened 2024-09-27
add benchmark for stumpless_open_buffer_target


goatshriek:
Buffer targets are use heavily in test suites, and will also be employed internally for new features. Adding a benchmark test for `stumpless_open_buffer_target` will give future changes a way to evaluate what the impact to performance is.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, read the [documentation](https://github.com/goatshriek/stumpless/blob/latest/docs/benchmark.md) on benchmarking in stumpless. This doc walks through the general philosophy behind benchmark tests and how they are used in this project.

Once you understand these concepts, you'll be ready to add a new benchmark suite as `test/performance/target/buffer.cpp` for `stumpless_open_buffer_target`. Look to the existing test suites as a starting point for your new one. To add the new benchmark suite to the build process, you'll need to modify `CMakeLists.txt` with a call to `add_performance_test` for the new benchmark suite. Make sure that you are able to see results from your new test when you run the benchmark suite using the `bench` make target.

m3g4d1v3r:
Hello there!
I've opened a pull request in order to solve this issue.
https://github.com/goatshriek/stumpless/pull/466
It's still a Work In Progress.

goatshriek:
Resolved by #466

PR #466, opened 2024-10-15
Add benchmark for stumpless_open_buffer_target

m3g4d1v3r:
This pull request is related to solving the issue for adding a new benchmark for stumpless_open_buffer_target. A new benchmark file was also added as described on the issue 449.

This is a Work in Progress: I hope to improve this progressively through feedback from the community.

### DIFF
--- a/test/performance/target/buffer.cpp
+++ b/test/performance/target/buffer.cpp
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+
+/*
+ * Copyright 2024 Joel E. Anderson
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <benchmark/benchmark.h>
+#include <stumpless.h>
+#include "test/helper/memory_counter.hpp"
+#define MAX_BUFFER_SIZE (4ull * 1024ull) // 4 kB
+
+NEW_MEMORY_COUNTER( buffer );
+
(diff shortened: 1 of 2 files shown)


m3g4d1v3r:
Issue #449

test/performance/target/buffer.cpp
```diff
@@ -0,0 +1,53 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <benchmark/benchmark.h>
+#include <stumpless.h>
+#include "test/helper/memory_counter.hpp"
+#define MAX_BUFFER_SIZE (1024 * 1024 * 1024 * 1) // 1 GB
```
goatshriek: Drop this by at least an order of magnitude so that it can run on smaller systems. Perhaps 4KiB?

goatshriek (changes requested):
Overall this looks great! Just one change requested to bring the resource usage down a bit.

test/performance/target/buffer.cpp
m3g4d1v3r: Hello, I'm sorry for the delay: I've got busy with personal stuff.

I've made a recent commit with the suggestion of setting this macro according to the user's cpu architecture. If you like this way, we can change the define into other magnitude values. Otherwise, we can rollback to defining it to 4KiB and keep it this way.

goatshriek: I think for just a test, just stick with a single hard-coded value; it's probably not worth trying to optimize for different architectures since this is just a test.

m3g4d1v3r: Nice, ok.
I've left as was suggested before: setting MAX_BUFFER_SIZE as 4KiB.

I hope to contribute more here in the future!

goatshriek: I will add that if you are interested in architecture-specific improvements don't let this discourage you: there is definitely a place for this in the library code base. The page size is one example of a place for improvement: if it can't be queried from the system using something like `sysconf` then it defaults to a single value: if you can find some better fallback values for `FALLBACK_PAGESIZE` in `CMakeLists.txt` for different architectures this could be a separate contribution.

goatshriek (approved):

goatshriek:
Thanks very much for putting this change together and sticking with it through the review process!

PR #468, opened 2024-10-20
add memory function accessors

tertyshnyi:
image
I added new 3 functions stumpless_get_malloc(), stumpless_get_free(), stumpless_get_realloc(). And added pointer and null tests to them. Is this enough?


tertyshnyi:
memory function accessors + tests

tertyshnyi:
I removed the function I mistakenly added earlier, it should work now.

goatshriek:
This is off to a great start, and I especially appreciate that you updated the test macros to use the new functions. A few things to help you get this over the finish line: don't forget to add the functions to `src/windows/stumpless.def`, and move the declarations in `include/stumpless/memory.h` within the `extern "C"` block so that the C++ name mangling isn't mistakenly applied to them.

tertyshnyi:
Thanks for the tip. I wrote functions in stumpless.def and wrote functions in extern “C” memory.h . I also changed the tests, removed the check for null because get functions do not return null, added 3 new ones that custom functions correctly replace standard functions and can be retrieved.

PR #469, opened 2024-10-21
have_sys_socket.c: MacOS support for sys_socket_sendto_udp_target() a…

vasiliyk:
Updated src/config/have_sys_socket.c to support MacOS.

MacOS does not support the flag MSG_NOSIGNAL used in:
- sys_socket_sendto_tcp_target()
- sys_socket_sendto_udp_target()  

but it has an equivalent SO_NOSIGPIPE that prevents SIGPIPE from being raised when a write fails on a socket.

### DIFF
--- a/src/config/have_sys_socket.c
+++ b/src/config/have_sys_socket.c
@@ -201,10 +201,21 @@ sys_socket_sendto_tcp_target( struct network_target *target,
       break;
     }
 
+/*
+ * MacOS does not support the flag MSG_NOSIGNAL but have an equivalent SO_NOSIGPIPE that prevents SIGPIPE from being
+ * raised when a write fails on a socket.
+ */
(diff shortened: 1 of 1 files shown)


goatshriek:
Thanks very much for submitting this, it isn't something I was aware of.

From what I can see there are other BSD systems (FreeBSD, for example) that also have this difference. I'd most prefer to add this as a configure check in the `CMakeLists.txt` to define a configuration `#define` for the socket option in either case. The process for this is described [here](https://github.com/goatshriek/stumpless/blob/latest/docs/style.md). This is a bit more work than what this pull request includes, so I understand if you'd rather not go through this. Let me know if you'd prefer not to go through that, and if so I'll put together a patch myself.

vasiliyk:
Joel, I will do it and open another PR.

Issue #443, opened 2024-09-26
add realloc_array function


goatshriek:
Stumpless includes wrappers for memory allocation to allow a customized memory allocator to be used. One of these is `realloc_mem`, which calls the underlying allocator's `realloc` routine.

As outlined in #440, there are situations where a multiplication is done to get the new size. For the same reasons outlined in that issue, it would also be helpful to add a `realloc` wrapper that does overflow checks.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

First, review #440 and understand the solution proposed there; this issue is very similar. **If you solved the `alloc_array` issue yourself, please leave this one for a new contributor. This is a "good first issue", and as such the learning experience is an essential component.**

Add a new function to `src/memory.c` named `realloc_array` with the following signature:

```c
void *
realloc_array( const void *mem, size_t item_count, size_t item_size );
```

Once you've made this addition, swap out the usage of `realloc_mem` with multiplication in `src/cache.c` `add_page` and `src/target/journald.c` `init_fields`, and make sure that all tests continue to pass.

Next, refactor  `stumpless_add_param` in `src/element.c` to use the new `realloc_array` function. You can then create a new test in `test/function/element.cpp` that modifies the `param_count` of the element structure to be sufficiently high (`SIZE_MAX` perhaps?) to cause an overflow and subsequent memory allocation failure to test your detection logic.

Brijeshthummar02:
i can fix this.

m3g4d1v3r:
Hello there!

I've created a pull request that aims to solve this issue: #472

ansh-dulewale:
I can fix this.

goatshriek:
Pull #472 already addresses this and is near completion; it would be better to find a different issue to work on.

goatshriek:
Resolved in #472.

PR #472, opened 2024-10-25
#443: Add realloc array function

m3g4d1v3r:
This pull request adds the realloc_array wrapper to the code base (it's inspired on alloc_array's implementation).
This is a Work In Progress. Please let me know if any fixes should be made.

Best regards.
m3g4d1v3r

### DIFF
--- a/src/element.c
+++ b/src/element.c
@@ -56,18 +56,13 @@ struct stumpless_element *
 stumpless_add_param( struct stumpless_element *element,
                      struct stumpless_param *param ) {
   struct stumpless_param **new_params;
-  size_t old_params_size;
-  size_t new_params_size;
 
   VALIDATE_ARG_NOT_NULL( element );
   VALIDATE_ARG_NOT_NULL( param );
 
   lock_element( element );
 
-  old_params_size = sizeof( param ) * element->param_count;
-  new_params_size = old_params_size + sizeof( param );
-
-  new_params = realloc_mem( element->params, new_params_size );
+  new_params = realloc_array( element->params, element->param_count + 1, sizeof( struct stumpless_param * ) );
   if( !new_params ) {
     unlock_element( element );
     return NULL;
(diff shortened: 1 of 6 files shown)


m3g4d1v3r:
Issue #443

src/memory.c
```diff
@@ -136,6 +136,16 @@ alloc_array( size_t item_count, size_t item_size ) {
   return alloc_mem(item_count * item_size);
 }
 
+void *
+realloc_array( const void *mem, size_t item_count, size_t item_size ) {
+  #include <stdio.h>
```
goatshriek: this include may be left over from testing? At any rate, if it is still needed it shouldn't be inside of this function.

src/element.c
```diff
@@ -56,18 +56,18 @@ struct stumpless_element *
 stumpless_add_param( struct stumpless_element *element,
                      struct stumpless_param *param ) {
   struct stumpless_param **new_params;
-  size_t old_params_size;
-  size_t new_params_size;
+  size_t old_params_count;
```
goatshriek: it appears that you can get rid of this local variable now, as the size calculation code that used it no longer needs it.

goatshriek (changes requested):
Thanks very much for putting this together! Just a few minor cleanups noted here and I believe it will be ready to merge.

src/memory.c
m3g4d1v3r: I'm sorry about this: I've really forgot to remove this header.

I've just refactored the 3 commits of this PR. Everything should be ok now.

src/element.c
m3g4d1v3r: Ok: I've removed these variables and the "+ 1" modification that went to new_params_size (let me know if this change is what is expected).

I've just refactored the 3 commits of this PR by doing git rebase.

goatshriek: The plus one is needed to reallocate the array with space for the new param, which is probably why some of the tests are failing. You can probably just use `element->param_count +1` as the second argument to `realloc_array` to fix this.

m3g4d1v3r: Hello! 

Right: I've just changed the second argument of this `realloc_array` call to `param_count + 1`.

goatshriek (approved):

Issue #335, opened 2023-04-10
migrate from libfuzzer to active framework


goatshriek:
libFuzzer has a note on its [website](https://llvm.org/docs/LibFuzzer.html#id13) stating that it is no longer under active work. Stumpless will need to migrate to an active fuzzing framework to avoid a rotten dependency. [FuzzTest](https://github.com/google/fuzztest) appears to be the successor to libFuzzer at the time of this writing.

jackoske:
not sure about rotten, to quote:
> The original authors of libFuzzer have stopped active work on it and switched to working on another fuzzing engine, [Centipede](https://github.com/google/centipede). LibFuzzer is still fully supported in that important bugs will get fixed. However, please do not expect major new features or code reviews, other than for bug fixes.

PR #474, opened 2024-10-28
refactor tests for invalid format items

vasiliyk:
refactor tests for invalid format items closing issue https://github.com/goatshriek/stumpless/issues/193


vasiliyk:
a few changes that I forgot earlier:
1. include moved form /test/helper/fixture.cpp to /include/test/helper/fixture.hpp
2. include added to stumpless_private.yml

vasiliyk:
@goatshriek, could you surmise why tests fail?

goatshriek:
I can't identify the exact cause from a quick glance. Are you seeing failures when you run the tests yourself?

There may be entries in the directory walk that you aren't expecting, based on some of the error messages being an "unknown file". Perhaps that would be the best place to start troubleshooting.

vasiliyk:
> Are you seeing failures when you run the tests yourself?

no, unfortunately, I have no errors on my Mac. I will add line terminators to the files that I added in directory ``stumpless/test/corpora/invalid_param_name``

vasiliyk:
ok, found it - dirent.h include is not available for Windows, I need to add code for Win additionally.

vasiliyk:
I made code Win and Unix/Linux/MacOS compatible and CLang pure (without C++ elements).
I am intrested to see Win/Linux tests, please allow tests execution.

vasiliyk:
Joel, some Windows tests are failing due to attempt to load 'dirent.h'. 

D:\a\stumpless\stumpless\test\helper\fixture.cpp(20,12): fatal error C1083: Cannot open include file: 'dirent.h': No such file or directory [D:\a\stumpless\stumpless\test_helper_fixture.vcxproj]

However, I have the following code:
```
#ifndef HAVE_WINDOWS_H
  #include <dirent.h>
#endif
```

Can it be a case where ``HAVE_WINDOWS_H`` is not defined for Windows test execution?

tools/check_headers/stumpless_private.yml
goatshriek, line 1: The changes in this file should point at the header the function lives in, not in which it was included in. For now, I think you can simply remove your changes to this file, and we can take another look at header changes when this is closer to done.

include/test/helper/fixture.hpp
```diff
@@ -85,4 +85,27 @@ create_nil_entry( void );
 const char *
 load_corpus( const std::string& name );
 
+struct stumpless_test_data {
+    int length;
+    char **test_strings;
+};
```
goatshriek: Use a C++ vector or other data structure here, instead of manually managing a list of strings this way.

test/function/param.cpp
```diff
@@ -235,19 +236,16 @@ namespace {
...
-    EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
-
-    result = stumpless_load_param( &param, "param]", "test-value" );
-    EXPECT_NULL( result );
-    EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
+    for(int i = 0; i < invalid_names.length; ++i) {
+      result = stumpless_load_param( &param, invalid_names.test_strings[i], "test-value" );
+      EXPECT_NULL( result );
+      EXPECT_ERROR_ID_EQ( STUMPLESS_INVALID_ENCODING );
+      delete[] invalid_names.test_strings[i];
```
goatshriek: As you refactor to use a C++ vector (or other structure), do your best to make it easier for these tests to run by not requiring manual deletion of strings.

goatshriek (changes requested):
First of all, thanks very much for putting together this change!

The `HAVE_WINDOWS_H` symbol is defined in the private configuration header, which is not exposed in the tests. This is why it is still being included in your tests, despite it being defined during the build itself. The header check in the Static Analysis would warn you about this.

Taking a step back though, it seems that filesystem operations in C++ can be hard to make portable, which isn't something I had anticipated.

Currently, the project does not impose any C++ version restrictions, instead simply inheriting from Google Test, which currently requires C++11. Newer versions of C++ (from C++17 onwards) have `<filesystem>`, which seems like it will make this much easier for you given this [Stack Overflow example](https://stackoverflow.com/questions/612097/how-can-i-get-the-list-of-files-in-a-directory-using-c-or-c).

Fortunately we're working on a major version release right now, so a major change like this is acceptable. Furthermore, I intend to migrate to [Google FuzzTest](https://github.com/google/fuzztest) as part of #335. FuzzTest requires C++17 anyway, so this standard will be in bounds for the C++ version required for testing. You may need to add a C++ standard statement to CMake for now to get things to build. If so you can add this anywhere, and I will refactor it when I update CMake and integrate FuzzTest later on.

## tl;dr
Refactor the file walk to use C++ `<filesystem>` which should resolve portability concerns and make your code much simpler.

I've made a few other suggestions in my review as well.

vasiliyk:
Joel, load_corpus_folder() logic converted to c++17 to simplify code

vasiliyk:
Based on error messages I guess the issue was in FS namespace for Windows.
I don't have Windows machine so I pushed the change to see the test results.

vasiliyk:
Joel, 
all Windows and MacOS builds passed tests, Linux builds on c++ passed as well.

What should be the next step?

goatshriek:
All of the tests need to pass. The Linux "with C++" tests are aimed at the C++ bindings, and are probably passing because they do not rely on the new code yet.

There is code in `tools/cmake/test.cmake` that sets the C++ standard - you may need to modify this to fix the builds. I recommend testing this in a container or virtual machine so that you can be sure it is fixed.

vasiliyk:
I set c++17 standard in ``tools/cmake/test.cmake`` in addition to ``CMakeList.txt``.

Great idea to use a container, I will setup it up once, and then it will speed-up testing for me.

vasiliyk:
Joel, 
there is a problem only with static analysis, I don't see the issue on my machine:

```
vasiliy@night stumpless % tools/check_headers/check_headers.rb "include/**/*.h*" "include/**/**/*.h"
warning: deprecated term stumpless_destroy_entry used in include/stumpless/entry.h
warning: deprecated term stumpless_destroy_element used in include/stumpless/element.h
warning: deprecated term stumpless_destroy_entry used in include/stumpless/entry.h
warning: deprecated term stumpless_destroy_element used in include/stumpless/element.h
```

Should I adjust ``stumpless_private.yml`` based on the details I see on github?

goatshriek:
The "warning" lines can be ignored. The cause for the failure is the messages below the warnings:

```
test/function/element.cpp: unused include test/helper/fixture.hpp
test/function/entry.cpp: 'string' requires inclusion of string
test/function/param.cpp: 'string' requires inclusion of string
test/function/param.cpp: unused include test/helper/fixture.hpp
test/helper/fixture.cpp: unused include filesystem
```

You should only need to add an entry to `tools/check_headers/cpp_standard_library.yml` for the `filesystem` include, but other than that you should be able to resolve the issues by taking the suggested action in each message.

vasiliyk:
headers configuration updated.

goatshriek (approved):

goatshriek:
Thank you once again for helping refactor the test suite, which is often overlooked. Your persistence in sticking with this through all of the requested modifications and portability concerns is especially commendable!

PR #457, opened 2024-10-03
Fixes #450: Adds new function for generating a shorter form prival string.

Jackb18b:
Adds a new function for a condensed and logger-standard prival string.

PR #476, opened 2024-11-07
450 stumpless get priority string

m3g4d1v3r:
This is a continuation from the other opened PR #457.

Since I don't have the permissions to commit to the other contributor's fork, I've decided to cherry pick them and publish a new PR.

I hope this is the proper way to do it.

Please, let me know if any adjustments are needed in this branch.

### DIFF
--- a/src/prival.c
+++ b/src/prival.c
@@ -58,6 +58,31 @@ stumpless_get_prival_string( int prival ) {
   return prival_string;
 }
 
+const char *
+stumpless_get_priority_string( int prival ) {
+  const char *severity;
+  const char *facility;
+  size_t priority_string_size;
+  char* priority_string;
+
+  severity = stumpless_get_severity_string( get_severity( prival ) );
+  facility = stumpless_get_facility_string( get_facility( prival ) );
+   
+  size_t len_severity = strlen(severity);
+  size_t len_facility = strlen(facility);
+
+  // +1 for '.' formatting, +1 for termination
+  priority_string_size = ( len_severity + len_facility + 2);
+  priority_string = alloc_mem( priority_string_size );
+
+  memcpy( priority_string, severity , len_severity); 
+  memcpy( priority_string + len_severity, ".", 1); 
+  memcpy( priority_string + len_severity + 1, facility, len_facility);
+  memcpy( priority_string + len_severity + 1 + len_facility, "\0", 1);
+
+  return priority_string;
+}
+
 int
 stumpless_prival_from_string( const char *string ) {
   int prival;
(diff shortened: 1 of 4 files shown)


goatshriek (changes requested):
Excellent work! There is one last change needed: please add a test that passes and invalid prival in, and add some error handling to ensure the response is sane. I personally think a NULL response in this case makes the most sense, but since the facility and severity string functions do not do this, I will defer to you for this change. Before the v3.0.0 release I will likely refactor all of them to return NULL in the event of an error so that the scenario is more clear.

The deadlock was not related to your changes, and is the second time I've seen it. I created #477 to capture the logs and remind myself to go back and look at this, but for now I only mention it to you as something you don't need to worry about.

m3g4d1v3r:
I've just added a commit with invalid privals checking and testing.

src/prival.c
```diff
@@ -58,6 +58,50 @@ stumpless_get_prival_string( int prival ) {
...
 
+const char *
+stumpless_get_priority_string( int prival ) {
+  const char *facility;
+  const char *severity;
+
+  size_t priority_string_size;
+  char* priority_string;
+
+  if ((prival & 0xff) != prival) {
```
goatshriek: You'll need to use the `facility_is_invalid` and `severity_is_invalid` functions to perform this check, since there are some values that are not valid privals despite fitting into a single byte.

goatshriek (changes requested):
Just a tweak on how an invalid prival is checked, and this should be good to go.

src/prival.c
m3g4d1v3r: Right! I've just amended the last commit and push forced the changes.

m3g4d1v3r: Just a second... The tests are failing

m3g4d1v3r: If I change the verification to:
`if ((facility_is_invalid(get_facility(prival))) || (severity_is_invalid(get_severity(prival)))) {`

then the negative integer test argument will always fail, since both getters mask the input.
Should I remove this test and keep the 0x100 one?

m3g4d1v3r: Apparently both -1 and 0x100 doesn't work in this case.

goatshriek: I would combine the test that you have with the invalid facility one. Thinking further, if you are testing for the lower byte mask, then the severity test will always pass since all 8 values are possible. The facility is the only one with additional things to worry about.

Also, it may be cleaner to factor this out so that the other prival functions can use it in the future.

goatshriek: It looks like `stumpless_set_entry_prival` suffers from the same bug you just pointed out. I will open a separate issue for that when I am at a terminal again.

m3g4d1v3r: I've just implemented a brief solution to this.
Tests are passing, things should work now.
I'm checking of byte range and also for facility range (the multiple of 8 wasn't necessary to test, since all masked values that are generated from `get_facility` are multiple of 8).

Let me know if any more corrections are needed.

goatshriek (approved):

goatshriek:
Thanks very much for finishing the resolution for this one!

Issue #225, opened 2022-03-21
automatic origin element


goatshriek:
[RFC 5424](https://datatracker.ietf.org/doc/html/rfc5424#section-7.2) specifies that a `origin` structured data element may be included to give some level of information about the originator of the log message. Stumpless should have the ability to generate this element automatically, as well as the capability to include it automatically when logging to given targets.

az1plo:
Hello, I'd like to work on this issue. Can you please assign me?

goatshriek:
Sure, you are welcome to work this one of you'd like.

vasiliyk:
Origin structure data element is basically 4 optional fields for Syslog Protocol: 

```
typedef struct {
    char ip[16];           // Origin IP address
    char enterpriseId[50]; // Enterprise ID
    char software[50];     // Software name
    char swVersion[20];    // Software version
} Origin;
```

It is clear what to put in 2 fields:
```
software = Stumpless
swVersion = stumpless-version 
 ```

``enterpriseId`` we probably need to skip.
``ip`` is open question as PC can have multiple network interfaces + multiple IPv4 together with IPv6 addresses. Should we make it configurable??

**Comment by goatshriek** [comment]
Yes, all of the fields in the origin element should be configurable.

The default for the IP param(s) could be either the first network address found when querying or all of the addresses on the machine; the RFC includes both options as something application MAY do. For the sake of simplicity, perhaps we start with the first address in this change, and update it later.

As you build this out, be sure that you're using the existing functionality where possible. Perhaps a new public function to get the origin `struct stumpless_element` instance, which the user can modify if they want to. You won't need to define a new struct for the origin element itself.

**Comment by vasiliyk** [comment]
Joel, a few questions:

1. where should public function be defined that returns an instance of ``stumpless_element`` with origin - ``log.c``?
2. where can I put helper functions like ``get_ip_addresses()`` - ``log.c`` jsut without ``STUMPLESS_PUBLIC_FUNCTION`` macro?
3. should I use C++ instead of C to avoid manual memory management as it complicates the code (see below)?

```
char** get_ip_addresses() {
  char **ip_addresses = NULL;
  int ip_addresses_counter = 0;
  char ip[INET6_ADDRSTRLEN];

  #ifdef _WIN32
    WSADATA wsaData;
    if ( WSAStartup( MAKEWORD(2, 2), &wsaData ) != 0 ) {
        perror( "WSAStartup failed" );
        return NULL;
    }

[...]

PR #479, opened 2024-11-17
draft: RFC5424 origin structure added

vasiliyk:
**draft: RFC5424 origin structure added; addressing #225**

What I don't like: 

1. default stumpless_element_to_string function returns the origin structure in an incorrect format. Correct format should be: ``[origin ip="example.ip" software="stumpless" swVersion="example.version"]``. Should I override function to account for it?
2. stumpless syslog and vsyslog replacements don't return origin structure by default. Should an option be added to them to include origin structure?

### DIFF
--- a/include/stumpless/log.h
+++ b/include/stumpless/log.h
@@ -689,4 +689,22 @@ vstumplog_trace( int priority,
 }                               /* extern "C" */
 #  endif
 
+/*
+ * Returns RFC 5424 origin structured data element that provides information about the originator of the log message.
+ * see: https://datatracker.ietf.org/doc/html/rfc5424#section-7.2
+ *
+ * It used by stumpless' syslog and vsyslogs replacement - stumplog, and vstumplog.
+ *
+ * @return stumpless_element with origin structure
+ * ip - Origin IP address (for now just first address used ipv4 or ipv6)
+ * enterpriseId - Enterprise ID (skipped for now)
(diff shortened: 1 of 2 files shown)


goatshriek (changes requested):
This will be quite a difficult addition, since there isn't an example/base set up for automatic elements yet.

The code to gather the IP addresses will need to be put into configuration source files and chosen based on the headers available, rather than the availability of WIN32. You can look at `include/private/config/wrapper/network_supported.h` to see the logic for this now. It would probably make the most sense to add to this, although the use of `ifaddrs.h` may warrant it's on header check and inclusion.

The automatic elements should be configurable, on by default, and not do too much special work apart from just creating and adding the element to entries before they're logged. For now, maybe adding a boolean flag to the target structure on whether or not automatic elements are enabled and then doing the extra work to add the element in `stumpless_add_entry`, or even better in `format_entry` is the best first step. This will allow you to bypass the concerns you note about the to_string format being incorrect, and the other syslog and vsyslog functions will still pick up these changes as well.

Also, be sure that you have some sort of test so that elements with the automatic element can be verified to have the correct data.

goatshriek:
Checking in to see if you plan to continue work on this, per the project's [inactivity policy](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md#you-still-there).

goatshriek:
Closing as stale.

Issue #446, opened 2024-09-26
remove unused strhelper functions


goatshriek:
The `src/strhelper.c` file has two functions, `copy_cstring_length` and `to_upper_case` that are no longer used. Remove these functions from the source file, associated header file `include/private/strhelper.h`, and header manifest in `tools/check_headers/stumpless.yml`. Of course, make sure that all builds and testing continue to succeed.

Brijeshthummar02:
i would like to work on this.

nhathuy13598:
Hi @Brijeshthummar02, can I work on this issue? Thanks

goatshriek:
Addressed in #486.

Sravz04086:
I would like to work on this issue.

PR #486, opened 2025-02-15
remove unused strhelper functions

jhayden02:
Regarding issue #446.

Removed strhelper function definitions from C files include/private/strhelper.h and src/strhelper.c, as well as from the header manifest tools/check_headers/stumpless.yml. Built successfully.

### DIFF
--- a/include/private/strhelper.h
+++ b/include/private/strhelper.h
@@ -34,12 +34,6 @@ copy_cstring( const char *str );
 char *
 copy_cstring_with_length( const char *str, size_t *length );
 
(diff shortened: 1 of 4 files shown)


goatshriek:
It looks like you've found that `ctype.h` is still needed to get `tolower` even though the header check tool didn't realize this. Please add a line to `tools/check_headers/c_standard_library.yml` for this to fix the unused header check when you add the include back.

goatshriek:
Thanks very much for putting this together!

Issue #80, opened 2020-07-15
implement ODELAY option


goatshriek:
Stumpless currently does not support the STUMPLESS_OPTION_ODELAY option despite providing it in its header files. This option is meant to mirror the LOG_ODELAY option in functionality. The meaning will change depending on the target - each target will need to be evaluated to make sure that it conforms to this default setting.

Don't forget to document the option after you implement it in `include/stumpless/option.h`.

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

iamsb97:
@goatshriek : Please assign this to me. Thanks!

goatshriek:
As with the NDELAY issue, please feel free to implement this one or two targets at a time, particularly since it will require some work for each one.

iamsb97:
I have raised a draft PR with 3 targets covered. WEL and SQLite3 might require significant refactoring. I am still trying to figure out the best way it can be done and will take it up in a subsequent PR. From the current targets covered, network and socket have no test added yet. I am unable to use network_target and socket_target structs in the tests as it throws a bunch of errors which seem to have something to do with C and C++ difference (not sure). Before I move ahead with figuring that out, I wanted to make sure the approach and code change so far looks okay.

goatshriek:
Resolved by #491.

PR #491, opened 2025-02-19
Implement ODELAY option

iamsb97:
Issue #80.

### Approach

Used an atomic variable to store default option and exposed two functions (setter and getter) to modify/retrieve the options. This allows initializing the target with certain options as ODELAY requires it be passed before target is created.

Benefits to this approach:

1. Ensures backward compatibility as nothing changes on the current user side behavior.
2. Initiates all subsequent targets with the same default options which helps avoid setting options for each target individually. It can still be overridden ofcourse.

Targets covered:

1. File
2. Network
3. Socket

Targets pending:

1. WEL
2. SQLite3

Targets not applicable:

1. Buffer
2. Stream
3. Journald
4. Chain

### DIFF
--- a/include/stumpless/target.h
+++ b/include/stumpless/target.h
@@ -892,11 +892,11 @@ stumpless_set_current_target( struct stumpless_target *target );
  *
  * **Async Signal Safety: AS-Safe**
  * This function is safe to call from signal handlers as it only consists of
- * an atomic read.
+ * an atomic write.
  *
  * **Async Cancel Safety: AC-Safe**
  * This function is safe to call from threads that may be asynchronously
- * cancelled, as it only consists of an atomic read.
+ * cancelled, as it only consists of an atomic write.
  *
  * @param option The bitwise 'or' of all options to be set as default.
  */
--- a/include/stumpless/target.h
+++ b/include/stumpless/target.h
@@ -491,6 +491,10 @@ stumpless_get_cons_stream( void );
  * to stumpless_set_current_target(), or the default target if neither of the
  * former exists.
  *
+ * The current target will still be the last target that was opened even if the
+ * target was opened with STUMPLESS_OPTION_ODELAY option and is not yet
+ * connected for logging.
+ *
  * If the target that is designated as the current target is closed, then the
  * current target will be reset to the default target until another target is
  * opened.
--- a/include/stumpless/target.h
+++ b/include/stumpless/target.h
@@ -519,6 +519,30 @@ struct stumpless_target *
 stumpless_get_current_target( void );
 
 /**
+ * Gets the options to be used while initializing a target.
+ *
+ * If no options are set for atomic variable default_option, it will default
+ * to 0.
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe. Atomic operations are used to work with the
+ * default option.
+ *
+ * **Async Signal Safety: AS-Safe**
+ * This function is safe to call from signal handlers as it only consists of
+ * an atomic read.
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled, as it only consists of an atomic read.
+ *
+ * @return Value store in atomic variable default_option
+ */
+STUMPLESS_PUBLIC_FUNCTION
+int
+stumpless_get_default_option( void );
+
+/**
  * Gets the default facility of a target.
  *
  * **Thread Safety: MT-Safe**
@@ -854,6 +878,31 @@ stumpless_set_cons_stream( FILE *stream );
 STUMPLESS_PUBLIC_FUNCTION
 void
 stumpless_set_current_target( struct stumpless_target *target );
+
+/**
+ * Sets the default options to be used for all new targets created.
+ *
+ * If a target is created without calling this function, STUMPLESS_OPTION_NONE 
+ * will be used as default. Using this function will not affect options for 
+ * targets already created.
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe. Atomic operations are used to work with the
+ * default target.
+ *
+ * **Async Signal Safety: AS-Safe**
+ * This function is safe to call from signal handlers as it only consists of
+ * an atomic read.
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled, as it only consists of an atomic read.
+ *
+ * @param option The bitwise 'or' of all options to be set as default.
+ */
+STUMPLESS_PUBLIC_FUNCTION
+void
+stumpless_set_default_option( int option );
 
 /**
  * Sets the default facility of a target.
--- a/include/stumpless/target.h
+++ b/include/stumpless/target.h
@@ -540,7 +540,7 @@ stumpless_get_current_target( void );
  */
 STUMPLESS_PUBLIC_FUNCTION
 int
-stumpless_get_default_option( void );
+stumpless_get_default_options( void );
 
 /**
  * Gets the default facility of a target.
@@ -902,7 +902,7 @@ stumpless_set_current_target( struct stumpless_target *target );
  */
 STUMPLESS_PUBLIC_FUNCTION
 void
-stumpless_set_default_option( int option );
+stumpless_set_default_options( int option );
 
 /**
  * Sets the default facility of a target.
--- a/src/target.c
+++ b/src/target.c
@@ -70,7 +70,7 @@
 static const char *target_type_enum_to_string[] = {
   STUMPLESS_FOREACH_TARGET_TYPE( GENERATE_STRING )
 };
-static config_atomic_int_t default_option = config_atomic_int_default;
+static config_atomic_int_t default_option = STUMPLESS_OPTION_NDELAY;
 static config_atomic_ptr_t current_target = config_atomic_ptr_initializer;
 static config_atomic_ptr_t default_target = config_atomic_ptr_initializer;
 static config_atomic_ptr_t cons_stream = config_atomic_ptr_initializer;
@@ -203,11 +203,9 @@ stumpless_add_entry( struct stumpless_target *target,
 
   options = target->options;
   if ( ( options & STUMPLESS_OPTION_ODELAY ) ) {
-    if ( !stumpless_target_is_open( target ) ) {
-      target = stumpless_open_target( target );
-      if ( !target ) {
-        return -1;
-      }
+    target = stumpless_open_target( target );
+    if ( !target ) {
+      return -1;
     }
   }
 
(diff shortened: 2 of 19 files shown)


src/target/file.c
```diff
@@ -60,9 +62,12 @@ stumpless_open_file_target( const char *name ) {
     goto fail;
   }
 
-  target->id = new_file_target( name );
-  if( !target->id ) {
-    goto fail_id;
+  options = stumpless_get_default_option(  );
+  if ( !((options & STUMPLESS_OPTION_ODELAY) && !(options & STUMPLESS_OPTION_NDELAY)) ) {
```
goatshriek: It's unfortunate that the Syslog API technically has these as different options, since they should be mutually exclusive. However, I don't think each call should separately check each flag. In this case, i would just check to see if NDELAY is set and create the file target if so.

Logic to make sure that the options flag is valid should probably just stay in the functions used to set the options themselves, such as `stumpless_set_default_option` and `stumpless_set_option`.

src/target/file.c
```diff
@@ -114,14 +119,26 @@ new_file_target( const char *filename ) {
...
 int
-sendto_file_target( struct file_target *target,
+sendto_file_target( struct stumpless_target *target,
                     const char *msg,
                     size_t msg_length ) {
   size_t fwrite_result;
+  struct file_target *ftarget;
+
+  if ( !target->id ) {
+    lock_target( target );
```
goatshriek: This lock should be acquired before reading the id, otherwise it may be inadvertently opened multiple times resulting in memory and handle leaks.

include/stumpless/target.h
```diff
@@ -518,6 +518,30 @@ STUMPLESS_PUBLIC_FUNCTION
...
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled, as it only consists of an atomic read.
+ *
+ * @return Value store in atomic variable default_option
+ */
+STUMPLESS_PUBLIC_FUNCTION
+int
+stumpless_get_default_option( void );
```
goatshriek: Make this function name plural `stumpless_get_default_options`, to reflect that multiple options may be set in the result.

include/stumpless/target.h
```diff
@@ -855,6 +879,31 @@ STUMPLESS_PUBLIC_FUNCTION
...
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled, as it only consists of an atomic read.
+ *
+ * @param option The bitwise 'or' of all options to be set as default.
+ */
+STUMPLESS_PUBLIC_FUNCTION
+void
+stumpless_set_default_option( int option );
```
goatshriek: Similarly, name this `stumpless_set_default_options`.

src/target.c
```diff
@@ -193,7 +194,7 @@ stumpless_add_entry( struct stumpless_target *target,
   VALIDATE_ARG_NOT_NULL_INT_RETURN( target );
   VALIDATE_ARG_NOT_NULL_INT_RETURN( entry );
 
-  if( unlikely( !target->id ) ) {
+  if( target->type != STUMPLESS_FILE_TARGET && unlikely( !target->id ) ) {
```
goatshriek: After this NULL check is where I would add the check of ODELAY, and subsequent calls to `stumpless_target_is_open` and `stumpless_open_target`.

goatshriek (changes requested):
This is off to a good start! I think scoping it to a few targets for the first implementation is a good way to start.

Rather than performing the checks and open calls in the `sendto` functions, I would prefer to see support added to `stumpless_open_target` and `stumpless_target_is_open` for these target types. `stumpless_add_entry` can then check the options on the target and these functions to see if `ODELAY` is set and attempt to open the target if appropriate.

This will allow users to have finer-grained control over precisely when a target is opened, using the same functionality that would need to be implemented anyway. If they don't want to tune this functionality, then it will happen automatically in `stumpless_add_entry` and they won't need to do anything other than set the option.

It'll also keep you from needing to make too many changes to the semantics of the sendto functions, like changing parameters, who holds what locks, etc.

iamsb97:
@goatshriek : I think I have fixed all the other errors except windows build. There aren't any details in workflow logs to understand what is going wrong. Can you help me with how to proceed here?

goatshriek:
From the build log it looks like the socket include added to `src/target.c` is causing an issue as the `sys/socket.h` is not available in Windows. The best way around this is probably to add some config wrappers to `include/private/config/wrapper/socket.h` for is open and open, similar to what the network targets use. The definitions of those functions would go in `src/target/socket.c` and be able to use the include with no issue, and the wrapper will prevent inclusion when support is not available.

iamsb97:
Thanks for your help and patience with this PR! I believe all workflow checks should pass now (maybe except coverage).

src/windows/stumpless.def
```diff
@@ -257,3 +257,5 @@ EXPORTS
   stumpless_get_realloc                         @232
   stumpless_get_priority_string                 @233
   stumpless_param_into_string                   @234
+  stumpless_set_default_option                  @235
+  stumpless_get_default_option                  @236
```
goatshriek: don't forget to apply the renaming here as well

goatshriek (changes requested):

src/windows/stumpless.def
iamsb97: Sure, thanks for pointing it out! I missed this one.

include/stumpless/target.h
```diff
@@ -855,6 +879,31 @@ STUMPLESS_PUBLIC_FUNCTION
...
+ * will be used as default. Using this function will not affect options for 
+ * targets already created.
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe. Atomic operations are used to work with the
+ * default target.
+ *
+ * **Async Signal Safety: AS-Safe**
+ * This function is safe to call from signal handlers as it only consists of
+ * an atomic read.
```
goatshriek: update to "atomic write"

include/stumpless/target.h
```diff
@@ -855,6 +879,31 @@ STUMPLESS_PUBLIC_FUNCTION
...
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * This function is safe to call from threads that may be asynchronously
+ * cancelled, as it only consists of an atomic read.
```
goatshriek: update to "atomic write"

src/target.c
```diff
@@ -69,6 +70,7 @@
 static const char *target_type_enum_to_string[] = {
   STUMPLESS_FOREACH_TARGET_TYPE( GENERATE_STRING )
 };
+static config_atomic_int_t default_option = config_atomic_int_default;
```
goatshriek: this should start as the default behavior, which is currently `STUMPLESS_OPTION_NDELAY`.

src/target.c
```diff
@@ -633,16 +651,43 @@ stumpless_open_target( struct stumpless_target *target ) {
...
 
   lock_target( target );
-  if( target->type != STUMPLESS_NETWORK_TARGET ) {
-    unlock_target( target );
-    raise_target_incompatible( L10N_TARGET_ALWAYS_OPEN_ERROR_MESSAGE );
-    return NULL;
-  }
+  switch ( target->type ) {
+    case STUMPLESS_FILE_TARGET:
+      ( ( struct file_target * )target->id )->stream = config_fopen( target->name, "a" );
```
goatshriek: I think there should be a check for whether the target is open yet, before opening a new handle to a target? Perhaps before this switch statement, a call to `stumpless_target_is_open` would more generically handle this.

src/target.c
```diff
@@ -633,16 +651,43 @@ stumpless_open_target( struct stumpless_target *target ) {
...
+      break;
+  }
+  
   unlock_target( target );
 
   return result;
+
+fail:
+  unlock_target( target );
+  stumpless_close_target( target );
```
goatshriek: Why close the target here? Would it not be better to leave the target state untouched in the case of a failure?

src/target.c
```diff
@@ -1098,8 +1162,11 @@ new_target( enum stumpless_target_type type, const char *name ) {
     goto fail_mutex;
   }
 
+  options = stumpless_get_default_options(  );
+
+  target->id = NULL;
   target->type = type;
-  target->options = STUMPLESS_OPTION_NONE;
+  target->options = STUMPLESS_OPTION_NONE | options;
```
goatshriek: this could just be a direct assignment of `stumpless_get_default_options`, no need for a local variable.

test/function/target.cpp
```diff
@@ -605,6 +605,13 @@ namespace {
     stumpless_free_all(  );
   }
 
+  TEST( GetDefaultOption, DefaultValue ) {
+    int option;
+
+    option = stumpless_get_default_options(  );
+    EXPECT_EQ( option, 0 );
```
goatshriek: You'll need to change this as well to match the default functionality of NDELAY.

test/function/target.cpp
```diff
@@ -1133,6 +1140,53 @@ namespace {
...
+                                           buffer,
+                                           sizeof( buffer ) );
+
+    option = stumpless_get_option( target, STUMPLESS_OPTION_ODELAY );
+    EXPECT_TRUE( option );
+
+    stumpless_close_buffer_target( target );
+    stumpless_free_all(  );
+
+    stumpless_set_default_options( STUMPLESS_OPTION_NONE );
```
goatshriek: it would probably be cleaner to get the default options before the call to set, and then use this result to reset the default options at the end. This way, if they change in the future this test won't need to be modified. I've marked the other places this should be done as well for update.

test/function/target/file.cpp
```diff
@@ -179,4 +179,48 @@ namespace {
...
+    EXPECT_NO_ERROR;
+    EXPECT_NOT_NULL( target->id );
+
+    stumpless_destroy_entry_and_contents( entry );
+    stumpless_close_file_target( target );
+
+    TestRFC5424File( filename, line_count );
+    remove( filename );
+
+    stumpless_set_default_options( STUMPLESS_OPTION_NONE );
```
goatshriek: same reset to original default options here.

test/function/target/network.cpp
```diff
@@ -331,6 +331,37 @@ namespace {
...
+      EXPECT_NULL( stumpless_target_is_open( target ) );
+
+      result = stumpless_add_log_str( target, 14, "sample-message" );
+      EXPECT_NOT_NULL( stumpless_target_is_open( target ) );
+
+      stumpless_close_network_target( target );
+
+    }
+
+    stumpless_set_default_options( STUMPLESS_OPTION_ODELAY );
```
goatshriek: same get and reset here

test/function/target/socket.cpp
```diff
@@ -212,6 +212,42 @@ namespace {
...
+    target = stumpless_open_socket_target( "socket-taken-test",
+                                           local_socket_name );
+    EXPECT_NOT_NULL( target );
+
+    result = stumpless_add_entry( target, entry );
+    EXPECT_LT( result, 0 );
+    EXPECT_EQ( stumpless_get_current_target(  ),
+               stumpless_get_default_target(  ) );
+
+    stumpless_set_default_options( STUMPLESS_OPTION_NONE );
```
goatshriek: same get and reset here

test/function/target/socket.cpp
```diff
@@ -229,6 +265,32 @@ namespace {
...
+    entry = create_entry(  );
+    EXPECT_NOT_NULL( entry );
+
+    result = stumpless_add_entry( target, entry );
+    EXPECT_NOT_NULL( stumpless_target_is_open( target ) );
+
+    stumpless_close_socket_target( target );
+    EXPECT_NO_ERROR;
+
+    stumpless_set_default_options( STUMPLESS_OPTION_NONE );
```
goatshriek: same get and reset here

test/function/target/file.cpp
```diff
@@ -179,4 +179,48 @@ namespace {
...
+    size_t i;
+
+    stumpless_set_default_options( STUMPLESS_OPTION_ODELAY );
+
+    target = stumpless_open_file_target( filename );
+    EXPECT_NO_ERROR;
+    EXPECT_NOT_NULL( target );
+    EXPECT_NOT_NULL( target->id );
+
+    entry = stumpless_new_entry( STUMPLESS_FACILITY_USER,
```
goatshriek: you can change this to the fixture `create_entry` function that is used in other places in this pull request

goatshriek (changes requested):
Great work on this! There are a number of tweaks outlined below, after which this should be good to merge.

iamsb97:
The line which I removed in this commit was added previously when I was closing the target on failure. Since close_target was removed, this assertion should have failed everytime but somehow it did not fail in linux and windows checks which seems a little strange to me.

goatshriek:
I'm not completely sure why that assertion was passing, as you're correct that it shouldn't. I will make a separate point to go look into that.

It does raise a point that probably needs clarification in the documentation though - that the current target will be set by an open call even if `ODELAY` is used on that target. As a final change to this, can you clarify this point in the documentation for the current target, which is the documentation for `stumpless_get_current_target` in `include/stumpless/target.h`. Adding a sentence to the block describing what the current target is should be enough.

goatshriek (approved):

goatshriek:
Thanks again for putting this together and sticking with the update through the requested changes! Excellent work.

Issue #183, opened 2021-06-08
Refactoring common portions of stumpless_param_to_string and stumpless_element_to_string


northernSage:
This is just a follow up issue to improve on #180. 

Quoting from #146:

> You may consider refactoring some common portions of `stumpless_param_to_string` into separate private functions that can be used in both places for efficiency...

Since #180 did not implement the changes described above I'll go ahead and work on a separate PR for this. I have opened the issue only to document my intent in doing so and to have somewhere to discuss related changes if the need arises.

goatshriek:
I'm adding more context to this issue so that it has enough information to be a viable good first issue for newcomers.

The common code between `stumpless_param_to_string` and `stumpless_element_to_string` may be tricky to identify, as it is not a simple case of looking at the two functions and finding similar code blocks. Instead, it is (1) the sizing of the resulting param string and (2) the writing of this string into a buffer. Both functions do this separately, and in the case of the element to string function this is redoing work that the param to string function already did.

One solution to this is to implement two new private functions that operate on already-locked parameters. One to get the needed size of the buffer for a parameter string, and another to write the string into a provided buffer and return the number of bytes written.

```c
// declared in include/private/param.h, defined in src/param.c:
size_t locked_get_param_string_size( struct stumpless_param *param );
size_t locked_param_into_buffer( struct stumpless_param *param, char *buffer, size_t buffer_size);
```

`stumpless_param_to_string` would then be changed to use these two functions to simply allocate a buffer and write into it. `stumpless_element_to_string` would also use them by: (1) locking all params to ensure they do not change between calculating the buffer size and writing to the buffer, (2) getting the sum of required sizes to calculate the total needed buffer size, (3) allocating this buffer, (4) writing the params into it, and (5) releasing the params.

This will cut down on heap memory allocations and avoid duplicate work, both of which will improve performance. Be sure to check the benchmarks of `stumpless_element_to_string` to verify that performance has actually improved. Refer to [docs/benchmark.md](https://github.com/goatshriek/stumpless/blob/latest/docs/benchmark.md) for details on how to use the benchmark suites.

As with all good first issues, there are a few details missing from this approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block.

Griezn:
I'll work on this issue when my other pr is merged, if no one has done this already.

ghost:
[...]

PR #493, opened 2025-04-08
Refactoring param and element functions

DevTigerart:
This PR addresses #183  and refactors stumpless_param_to_string, stumpless_param_into_string and stumpless_element_to_string so that they use mutualized code.

### DIFF
--- a/src/param.c
+++ b/src/param.c
@@ -286,14 +286,14 @@ stumpless_param_to_string( const struct stumpless_param *param ) {
 
     len = locked_get_param_string_size( param );
 
+    unlock_param( param );
+
     format = alloc_mem( len );
     if( !format ) {
       goto fail;
     }
     
-    (void)locked_param_into_buffer( param, format, len );
-
-    unlock_param( param );
+    (void)stumpless_param_into_string( param, format, len );
 
     clear_error( );
     return format;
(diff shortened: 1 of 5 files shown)


goatshriek:
You can ignore the Cygwin failure. I'll take a more in depth look at this when I have more time, for now it seems like there is just an included issue to fix.

src/param.c
```diff
@@ -278,37 +278,22 @@ stumpless_set_param_value( struct stumpless_param *param, const char *value ) {
...
-
-    unlock_param( param );
-
-    format[name_len ] = '=';
-    format[name_len + 1] = '\"';
-    format[name_len + value_len + 2] = '\"';
-    format[name_len + value_len + 3] = '\0';
-
+    
+    (void)stumpless_param_into_string( param, format, len );
```
goatshriek: Hmm, now that I'm looking closer at this, I think you'll need to extend the param lock to cover this call, and change this to use `locked_param_into_buffer` instead. Otherwise, it's possible for the param to change between the measurement and the allocation, which could lead to memory corruption. Sorry for the back and forth on this.

src/param.c
```diff
@@ -278,37 +278,22 @@ stumpless_set_param_value( struct stumpless_param *param, const char *value ) {
...
-    value_len = param->value_length;
+    len = locked_get_param_string_size( param );
 
-    /* name="value"*/
-    format = alloc_mem( value_len + name_len + 4 );
+    unlock_param( param );
+
+    format = alloc_mem( len );
     if( !format ) {
       goto fail;
```
goatshriek, line 291: note that this would lead to a double-unlock with the unlock being called before it, as the fail branch also unlocks the param. Extending the lock as described below should resolve this.

goatshriek (changes requested):
This looks good! Only a small change requested, mostly because I steered you down an incorrect path in the first place, sorry about that.

src/param.c
DevTigerart: No problem I completely missed the unprotected area. Will change that.

src/param.c
DevTigerart: Oopsy, missed that too ! Also I realized that I left unused variables in stumpless_param_into_string so I also fixed that.

DevTigerart:
I also realized I forgot to update the changelog. It's now done !

Issue #494, opened 2025-04-09
Update french (fr-fr) localization


DevTigerart:
Hi,

I will update the french translation.

Best regards

PR #495, opened 2025-04-09
Update fr-fr localization

DevTigerart:
This PR addresses #494 and updates french localization.


goatshriek:
Thanks very much for updating this translation, the language support is entirely community supported and not something I can provide myself.

If you enjoy adding language support and would like to improve the accessibility for French speakers, consider adding a `fr-fr` directory to the root  `l10n` with a translation of the top part of the README, as korean and chinese already have. You can then link this from the main README, and this can get expanded later on as other folks with the skills and time come through.

Issue #448, opened 2024-09-26
entry to string function


goatshriek:
It is often useful to be able to obtain a string representation of a given value either for user interface hints, troubleshooting, or operations like value comparison. Such a function is currently implemented for some stumpless data types (like the version, param, and element structs) but not for others, including the entry struct.

Note that this function is _not_ intended for use in the formatting of a log message as this output may be subtly different, for example when escape codes are needed. Rather, it is meant to give an as-is representation of the entry itself (all fields and elements).

Check out the [Contributing Guidelines](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md) and the [development guide](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md) for the basics on working with stumpless and submitting changes.

## General Approach
There are a few details left out of the following approach, for you to fill in as you encounter them. If you find you need help, please ask here or on the [project gitter](https://gitter.im/stumpless/community) and someone can help you get past the stumbling block. You are also free to take a different approach if you feel it is better - this is merely meant as a starting point.

First, read [this specific section](https://github.com/goatshriek/stumpless/blob/latest/docs/development.md#adding-new-functions) of the development guide devoted to adding new functions. It provides an outline of the work you will need to do beyond the implementation and testing of the function.

Next, add a declaration for your function in the `include/stumpless/entry.h` header file. Name the function `stumpless_entry_to_string` and document its behavior in the same doxygen format as the others in the same header. Make sure to cover the following points:
 * the general format of the output. This should be each field name (procid, app_name, msgid, etc.), an equals sign, and the value in double quotes, all separated by commas. For the elements, use the `stumpless_element_to_string` function to get the output for each element. For example: `time="2003-10-11T22:14:15.003Z", prival="165", hostname="example.com", app_name="test-suite", procid="6", msgid="yellow", element-1=[param="val"], message="test message"`
 * the supplied argument should be a `const` pointer
 * explicitly state that the result must be freed when it is no longer needed

[...]

PR #501, opened 2025-05-11
#448 - Added stumpless_entry_to_string and test cases for output format and empty fields

junyeong0619:
#448

I added stumpless_entry_to_string and a private function append_key_value_pair to reduce the amount of repetitive code in the main function. I also added test cases for OutputFormatCheck, NullEntry, and cases with empty hostname and procid.

P.S.
This is my first pull request. If there’s anything wrong with my code, please let me know. I sincerely appreciate your feedback. Thank you!


goatshriek (changes requested):
For some reason the CI runs weren't triggered on the latest commit, but nonetheless there are several remaining issues for you to work out. Please make sure that you're testing your changes locally, or by enabling Github Actions on your own fork, before pushing to this PR. The CI checks for pull requests are too slow to use them to test each change, largely because they need manual approval to run.

Make sure you read through the documentation for new functions. This outlines several steps that are still missing here:

 * add the new function to `include/stumpless/entry.h` with the `STUMPLESS_PUBLIC_FUNCTION` macro.
 * add an entry to `src/windows/stumpless.def` for the new function
 * add an entry to `tools/check_headers/stumpless.yml` for the new function to pass header checks

Also, make sure that you use `alloc_mem` instead of `malloc` directly, as the former allows the memory allocator to be changed, and is how some of the tests around memory allocation work.

junyeong0619:
> For some reason the CI runs weren't triggered on the latest commit, but nonetheless there are several remaining issues for you to work out. Please make sure that you're testing your changes locally, or by enabling Github Actions on your own fork, before pushing to this PR. The CI checks for pull requests are too slow to use them to test each change, largely because they need manual approval to run.
> 
> Make sure you read through the documentation for new functions. This outlines several steps that are still missing here:
> 
> * add the new function to `include/stumpless/entry.h` with the `STUMPLESS_PUBLIC_FUNCTION` macro.
> * add an entry to `src/windows/stumpless.def` for the new function
> * add an entry to `tools/check_headers/stumpless.yml` for the new function to pass header checks
> 
> Also, make sure that you use `alloc_mem` instead of `malloc` directly, as the former allows the memory allocator to be changed, and is how some of the tests around memory allocation work.

Thank you for your advice. I wil try this right now.

goatshriek:
The current failures are related to the tests not having enough coverage, and a memory leak being detected. Be sure that you are freeing all the memory you allocate in your tests, as this can cause a valgrind failure.

junyeong0619:
> The current failures are related to the tests not having enough coverage, and a memory leak being detected. Be sure that you are freeing all the memory you allocate in your tests, as this can cause a valgrind failure.

thank you i will check it right now

junyeong0619:
> The current failures are related to the tests not having enough coverage, and a memory leak being detected. Be sure that you are freeing all the memory you allocate in your tests, as this can cause a valgrind failure.

I found out that char * of element_to_string didn't free. Please check again I'm really sorry.

include/private/entry.h
```diff
@@ -27,6 +27,8 @@
 #include <stumpless/severity.h>
 #include "private/strbuilder.h"
 
+
+
```
goatshriek: Remove the changes to this file so that it isn't included in this change.

include/stumpless/entry.h
```diff
@@ -1818,3 +1826,5 @@ vstumpless_set_entry_message( struct stumpless_entry *entry,
 }                               /* extern "C" */
 #  endif
 #endif                          /* __STUMPLESS_ENTRY_H */
+
+
```
goatshriek: Remove this whitespace addition from your change.

include/stumpless/entry.h
```diff
@@ -26,6 +26,7 @@
 #  include <stdarg.h>
 #  include <stdbool.h>
 #  include <stddef.h>
+#  include <stdlib.h>
```
goatshriek: This header addition does not seem relevant to your change, you should be able to remove it.

src/entry.c
```diff
@@ -1483,3 +1484,113 @@ void
 unlock_entry( const struct stumpless_entry *entry ) {
   config_unlock_mutex( entry->mutex );
 }
+
+
+/*
+ * formating entry_to_string values
+ */
+static size_t append_key_value_pair(char *dest, size_t offset, const char *key, const char *value, size_t value_length) {
+    size_t key_len = strlen(key);
```
goatshriek: It would be safer to pass in a key size parameter, the same as you've done with the value parameter, and avoid this call to `strlen`.

src/entry.c
```diff
@@ -1483,3 +1484,113 @@ void
 unlock_entry( const struct stumpless_entry *entry ) {
   config_unlock_mutex( entry->mutex );
 }
+
+
+/*
+ * formating entry_to_string values
```
goatshriek: Complete the documentation of this function as well, to make it easier to use elsewhere in the future.

src/entry.c
```diff
@@ -1483,3 +1484,113 @@ void
...
+
+    // "
+    dest[offset++] = '"';
+    dest[offset++] = ',';
+
+    return offset;
+}
+
+char *stumpless_entry_to_string(const struct stumpless_entry *entry) {
+    if (!entry) {
```
goatshriek: Use `VALIDATE_ARG_NOT_NULL` for this check, so that the error code is set appropriately.

src/entry.c
```diff
@@ -1483,3 +1484,113 @@ void
...
+        return NULL;
+    }
+
```
goatshriek, line 1518: You'll need to obtain the lock on the entry before gathering these values (and be sure to release it at the end as well). Otherwise, this will not be thread-safe code.

src/entry.c
```diff
@@ -1483,3 +1484,113 @@ void
...
+      size_t len = strlen(element_str);
+      if (len > 0) {
+        if (!is_first_element) {
+          return_format[offset++] = ',';
+        }
+        memcpy(return_format + offset, element_str, len);
+        offset += len;
+        is_first_element = 0;
+      }
+      free((void *)element_str);
```
goatshriek: This should be a call to `free_mem`, to match `alloc_mem` usage within the library. See the `memory.h` header for details on this.

src/windows/stumpless.def
```diff
@@ -259,3 +259,4 @@ EXPORTS
   stumpless_param_into_string                   @234
   stumpless_set_default_options                 @235
   stumpless_get_default_options                 @236
+  stumpless_entry_to_string                           @237
```
goatshriek: align the ordinal number with the others (column-wise)

test/function/entry.cpp
```diff
@@ -67,27 +67,27 @@ namespace {
...
+      element_2 = stumpless_new_element( element_2_name );
+      stumpless_add_element( basic_entry, element_2 );
 
-        nil_entry = create_nil_entry();
+      nil_entry = create_nil_entry();
 
-        
-      }
+
+    }
```
goatshriek: For some reason the formatting has been changed here - please revert it so that it is not included in the change.

goatshriek (changes requested):
I've gone ahead and provided an initial review for you, though I typically wait until the tests are all passing. This cleanup should help you find the source of the valgrind error as you work through things.

Be sure that as you work, you look at the code around what you're adding to see how things like parameter validation, memory allocation, error handling, locking etc. are done. This will help you avoid pitfalls are already handled elsewhere.

test/function/entry.cpp
junyeong0619: I rebased all of the commits before. how can i erase that? i handled other comments

goatshriek: If you add another commit undoing this, it will be removed from the final commit.

In general, you don't need to worry about rebasing on this project as you go, since pull requests are squashed on merge anyway.

junyeong0619: i think i maybe resolved this problem. I tried to delete and download your original entry.cpp and add my new testcodes in it and push commit. I'm really sorry to spend your time.

junyeong0619: change free->free_mem inside the testcode(wrong change)

junyeong0619:
I handled entry.c 1548 line valgrind problem.

goatshriek:
It looks like you're almost there! I will give this a more thorough review when I have time to sit down with it, and after that you should be all set after any changes from that.

junyeong0619:
> It looks like you're almost there! I will give this a more thorough review when I have time to sit down with it, and after that you should be all set after any changes from that.

In codecov partial problem is lack of test?

goatshriek:
Yes. Some of the conditions you cannot hit and can be safely removed, and others you will need to orchestrate a memory failure test for. I can give more detail on those once I look at it in more detail.

junyeong0619:
> Yes. Some of the conditions you cannot hit and can be safely removed, and others you will need to orchestrate a memory failure test for. I can give more detail on those once I look at it in more detail.

Before I go to bed, i changed some codes myself . I hope it wil work well.

junyeong0619:
I think i made a test, but it reference the basic_entry and set element's pointer to nullptr. So the valgrind error occured. In new commit I copy the basic_entry and set the nullptr value to element.

junyeong0619:
i tried to remove the lines that have partial error in codecov and it works well in my tests. Will i update commit on this version?
Furthermore, I wonder how can I check codecov system in my local.

goatshriek:
You can find instructions on locally generating coverage reports in the [testing docs](https://github.com/goatshriek/stumpless/blob/latest/docs/test.md#test-coverage).

Hold off on making any more changes, and I will give you a detailed review with final steps. However, the documentation for the function in `stumpless/entry.h` is still missing, this would be good for you to work on in the meantime, along with removing unrelated whitespace changes from your changes.

junyeong0619:
> You can find instructions on locally generating coverage reports in the [testing docs](https://github.com/goatshriek/stumpless/blob/latest/docs/test.md#test-coverage).
> 
> Hold off on making any more changes, and I will give you a detailed review with final steps. However, the documentation for the function in `stumpless/entry.h` is still missing, this would be good for you to work on in the meantime, along with removing unrelated whitespace changes from your changes.

I understand what you saying. I will check empty space and add description on entry.h

include/private/entry.h
goatshriek, line 1: Remove this change from the pull request. You can either add the deleted line back, or simply copy this file from the default branch to yours.

include/stumpless/entry.h
```diff
@@ -45,6 +45,7 @@
 /** The maximum length of a msgid, as specified by RFC 5424. */
 #  define STUMPLESS_MAX_MSGID_LENGTH 32
 
+
```
goatshriek: Remove this unneeded whitespace change.

include/stumpless/entry.h
```diff
@@ -123,6 +124,34 @@ struct stumpless_entry {
...
+  * This function is not safe to call from threads that may be asynchronously
+  * cancelled, due to the use of a lock that could be left locked as well as
+  * memory management functions.
+  *
+  *
+  * @param entry The entry to get all elements
+  *
+  *@return The formatted string of each entry's element  will formatted by element="value"
+  *and each string attach by " , ". But entry's struct stumpless_element section is made by
+  * stumpless_element_to_string() and attach.
```
goatshriek: Make sure that the line lengths are at most 80 characters long.

src/entry.c
```diff
@@ -1483,3 +1484,121 @@ void
 unlock_entry( const struct stumpless_entry *entry ) {
   config_unlock_mutex( entry->mutex );
 }
+
+
+
+
+static size_t append_key_value_pair(char *dest, size_t offset, const char *key, size_t key_length, const char *value, size_t value_length) {
```
goatshriek, line 1495: Add a documentation comment above the function definition describing what it does.

src/entry.c
```diff
@@ -1483,3 +1484,121 @@ void
...
+
+  return offset;
+}
+
+/*
+ * format and print struct entry to element_name="value"
+ * to format struct entry's elements used stumpless_element_to_string function
+ * if new element of struct entry created it should be added new element's string lines
+ * it use thread_safe function
+ */
```
goatshriek: You can remove this block comment - the documentation in `stumpless/entry.h` is the correct place for this information, and it is already covered there.

src/entry.c
```diff
@@ -1483,3 +1484,121 @@ void
...
+  offset = append_key_value_pair(return_format, offset, "message_id", strlen("message_id"), message_id, message_id_length);
+
+  // process_id
+  offset = append_key_value_pair(return_format, offset, "process_id", strlen("process_id"), process_id, process_id_length);
+
+  // message
+  offset = append_key_value_pair(return_format, offset, "message", strlen("message"), message, message_length);
+
+
+  int is_first_element = 1;
```
goatshriek: use a `bool` type for this variable instead of an integer, to more clearly convey the intent

test/function/entry.cpp
```diff
@@ -23,6 +23,7 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 #include <stumpless.h>
+
```
goatshriek: Remove this unneeded whitespace change.

test/function/entry.cpp
```diff
@@ -97,6 +98,76 @@ namespace {
...
+    char *result = stumpless_entry_to_string(basic_entry);
+    ASSERT_NE(result, nullptr);
+    EXPECT_TRUE(strstr(result, "hostname=\"\"") != nullptr);
+    EXPECT_TRUE(strstr(result, "process_id=\"\"") != nullptr);
+    EXPECT_TRUE(strstr(result, "message_id=\"basic-msgid\"") != nullptr);
+    free(result);
+  }
+
+  TEST_F(EntryTest, NoneOfElementCheck) {
+    struct stumpless_entry test_entry = *basic_entry;
```
goatshriek: Instead of `basic_entry`, use `create_empty_entry()`, which will create an entry without any elements. You could either add this to the fixture setup/teardown, or just call the function in this test. This will avoid depending on struct copy mechanics and internal fields of the entry struct. The `nil_entry` fixture is also an option for this test I believe, which is already handled by the fixture setup.

goatshriek (changes requested):
Thanks very much for sticking with this all the way through, you're almost there! Most of these changes are cosmetic, with just a few substantial changes that shouldn't cause too much trouble for you.

junyeong0619:
I hope the changes I made are correct. I appreciate your help as always.

src/entry.c
```diff
@@ -1483,3 +1484,119 @@ void
...
+    size_t hostname_length = entry->hostname_length;
+    char *message = entry->message;
+    size_t message_length = entry->message_length;
+    const char *message_id = entry->msgid;
+    size_t message_id_length = entry->msgid_length;
+    const char *process_id = entry->procid;
+    size_t process_id_length = entry->procid_length;
+    size_t entry_priority_value = entry->prival;
+
+    unlock_entry(entry);
```
goatshriek: This unlock needs to be moved to after the fields of the entry are actually read. The pointers for things like the procid could change if the lock is not held, which would result in a read of memory that is potentially no longer valid.

src/entry.c
```diff
@@ -1483,3 +1484,119 @@ void
...
+      size_t len = strlen(element_str);
+        if (!is_first_element) {
+          return_format[offset++] = ',';
+        }
+        memcpy(return_format + offset, element_str, len);
+        offset += len;
+        is_first_element = false;
+      free_mem(element_str);
+  }
+
```
goatshriek, line 1592: Moving the unlock here should be safe - this will also keep the elements from being changed during the loop to read them.

goatshriek (changes requested):
Excellent work! There's just one more change for thread safety that I missed on the last time around (my apologies), and then this will be good to go.

junyeong0619:
I understood why the unlock should be in that position. I got learning a lot of things from you.

goatshriek (approved):

goatshriek:
It was great to watch this implementation grow into a well-formed and robust enhancement to the library. Thanks very much for putting this together, and even more so for sticking with it through all of the revisions.

PR #502, opened 2025-05-17
korean translate for two defines

junyeong0619:
I edit for two todo translate for korean.


goatshriek:
Thanks very much for putting this together! Translation contributions are always greatly appreciated.

PR #503, opened 2025-05-27
refactored config_files.cmake

### LINKED ISSUE #500: refactor level config header generation
Stumpless uses CMake for the build of the project, including the generation of configuration headers. One of these steps, in `tools/cmake/config_files.cmake`, generates the header files for the logging levels (alert, info, notice, etc.). This code is currently a series of copy-pasted calls to `configure_file`, which could be refactored into a single loop across a few lists with the appropriate variables. This task is to refactor this portion of the CMake build script into a loop so that it is clearer what is happening.

## General Approach

You can write a loop in CMake using one of the [foreach](https://cmake.org/cmake/help/latest/command/foreach.html) commands. A straightforward solution would be to have two lists, one for the long name, and one for the shortname. The loop could iterate over the range of the list, indexing into each one to set the variables before the call to `configure_file`. The [`string`](https://cmake.org/cmake/help/latest/command/string.html) commands will also be useful, particularly `TOUPPER` and `TOLOWER` operations to build the upper and lower case names.

Once you have this refactored, make sure that the library continues to build and all tests continue to pass. You may need to delete and re-create the build directory periodically, to make sure that an old version of the script has not left the build in an mixed up state.

itsomort:
Closes #500. 
Refactored config_files.cmake to use a list for generating header files of different logging levels. This cuts down on copy-pasted code and allows for an easier time making new logging levels in the future.
This builds successfully, and compiling with `--target check`, all test suites pass.

### DIFF
--- a/tools/cmake/config_files.cmake
+++ b/tools/cmake/config_files.cmake
@@ -1,15 +1,18 @@
 configure_file(
-  "${PROJECT_SOURCE_DIR}/include/stumpless/config.h.in"
-  "${PROJECT_BINARY_DIR}/include/stumpless/config.h"
+    "${PROJECT_SOURCE_DIR}/include/stumpless/config.h.in"
+    "${PROJECT_BINARY_DIR}/include/stumpess/config.h"
 )
+
 configure_file(
-  "${PROJECT_SOURCE_DIR}/include/private/config.h.in"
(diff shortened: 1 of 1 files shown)


itsomort:
I made a typo

PR #504, opened 2025-05-27
config_files.cmake refactoring

### LINKED ISSUE #500: refactor level config header generation

itsomort:
Closes #500.
Refactored config_files.cmake to use a list for generating header files of different logging levels. This cuts down on copy-pasted code and allows for an easier time making new logging levels in the future.
This builds successfully, and compiling with --target check, all test suites pass. This second pull request is because I made a small typo without realizing. First time contributor bad luck it seems.


itsomort:
Checking back on this, I'm not too sure why the checks aren't successful. It's showing that the stumpless/level/*.h files aren't being generated for some reason. I'm not too familiar with GitHub Actions and automatic checks, so I suspect something is messed up on my end. Cloning my fork of the repo and running `cmake --build . --target check` shows no failing tests.

goatshriek:
In addition to a fresh clone, use an out of tree build to make sure leftover files aren't still present. Something like:

```
mkdir /tmp/build
cd /tmp/build
cmake /path/to/stumpless
make all
```

itsomort:
Even with a fresh clone and making a new build directory, I'm getting no errors when compiling and running.
The exact commands I run:
```
mkdir test2
cd test2
git clone https://github.com/itsomort/stumpless.git
mkdir build
cd build
cmake ../stumpless
cmake --build . --parallel 4 --target check
```
After this, it gives 100% tests passed, 0 tests failed out of 63. However, going into the `include/stumpless/level` directory, no files appear. So the tests are passing without the include files actually being created. I'm checking why the include files aren't being created, but I was mislead by the tests passing while the files weren't being created.

itsomort:
I misread the cmake documentation for `ZIP_LISTS` and wrote `in` rather than `IN`. Honestly looking back, I'm not too sure why I thought this would be the thing to be lowercase out of everything, but it is what it is. The .h files for all the levels are under `include/stumpless/level`, so it should be fixed now.

goatshriek:
I'm not sure why the tests were still passing for you, did you inadvertently install the library at some point? That would make the headers available even when the build did not generate them.

Looks like the tests are passing now, which is great! I will provide a review with some change requests in the next day or so, when I have time to sit down and write them out.

itsomort:
> I'm not sure why the tests were still passing for you, did you inadvertently install the library at some point? That would make the headers available even when the build did not generate them.
> 
> Looks like the tests are passing now, which is great! I will provide a review with some change requests in the next day or so, when I have time to sit down and write them out.

It was definitely installing the library at some point. Thanks for the help!

tools/cmake/config_files.cmake
```diff
@@ -22,83 +25,17 @@ configure_file(
...
-
-set(LEVEL_NAME_UPCASE "NOTICE")
-set(LEVEL_NAME_DOWNCASE "notice")
-set(LEVEL_SHORTNAME "n")
-configure_file(
-  "${PROJECT_SOURCE_DIR}/include/stumpless/level/level.h.in"
-  "${PROJECT_BINARY_DIR}/include/stumpless/level/${LEVEL_NAME_DOWNCASE}.h"
-  @ONLY
-)
+set(NAME_UPCASE ALERT CRIT DEBUG EMERG ERR INFO NOTICE WARNING TRACE)
```
goatshriek: Get rid of this list, so that changes only need to happen in the other two.

tools/cmake/config_files.cmake
```diff
@@ -22,83 +25,17 @@ configure_file(
...
+set(NAME_DOWNCASE alert crit debug emerg err info notice warning trace)
+set(NAME_SHORT a c d em er i n w t)
 
-set(LEVEL_NAME_UPCASE "WARNING")
-set(LEVEL_NAME_DOWNCASE "warning")
-set(LEVEL_SHORTNAME "w")
-configure_file(
+foreach(UPCASE DOWNCASE SHORTNAME IN ZIP_LISTS NAME_UPCASE NAME_DOWNCASE NAME_SHORT)
+  set(LEVEL_NAME_UPCASE "${UPCASE}")
```
goatshriek: Use `string`'s [`TOUPPER`](https://cmake.org/cmake/help/latest/command/string.html#toupper) command to set the UPCASE name instead.

tools/cmake/config_files.cmake
```diff
@@ -22,83 +25,17 @@ configure_file(
...
+  set(LEVEL_NAME_DOWNCASE "${DOWNCASE}")
+  set(LEVEL_SHORTNAME "${SHORTNAME}")
```
goatshriek: Can you omit these two `set` invocations by changing the name of the variables in the `foreach` to match these instead?

goatshriek (changes requested):
This is exactly what I was hoping for in this refactoring, well done! Just a few more tweaks to clean things up, and then this should be good to go.

itsomort:
All changes from the review request should be implemented. Let me know if there's anything else that needs to be altered.

goatshriek:
Thanks very much for putting this together, and sticking with it through the issues and change requests!

PR #505, opened 2025-06-09
feat: Add facility function tests and fix missing test targets

### LINKED ISSUE #456: add errors to facility functions
The functions dealing with facilities do not currently properly handle error codes. They should generate appropriate error messages when a failure occurs, so that users checking for errors are properly notified. Additionally, errors should be cleared in the event of success so that an error is not mistakenly identified for a successful call. The affected functions are:

 * `stumpless_get_facility_string`
 * `stumpless_get_facility_enum`
 * `stumpless_get_facility_enum_from_buffer`

## General Approach

Add your new tests to the test suite for the facility functions `test/function/facility.cpp`. Each function should have a new test for error clearing, and the existing tests should be enhanced.

The new tests will cause a failure by misusing a different function in order to purposefully generate an error. `stumpless_version_to_string` is a simple function to misuse (pass it a NULL arg) as one candidate. Assert that there is in fact an error with `EXPECT_ERROR_ID_EQ`. It will then call the facility function normally, check the result, and ensure that no error is raised with `EXPECT_NO_ERROR`. Look to other tests for usage of these expect macros. Ensure that the new tests fail, and then add calls to `clear_error` in the functions themselves to fix the issue.

For existing tests, add calls to `EXPECT_ERROR_ID_EQ` with `STUMPLESS_INVALID_FACILITY` as the expected error. Update the functions themselves to raise this error upon failure, and update the documentation to reflect this behavior.

rs0125:
### Improve Facility Error Handling (Fixes https://github.com/goatshriek/stumpless/issues/456)

This PR improves error handling and test coverage for the facility-related functions in the library. It also introduces a new function test target to aid future debugging.

---

#### `src/facility.c`

* **`stumpless_get_facility_string`**:

  * Now calls `clear_error()` on success.
  * Calls `raise_error(STUMPLESS_INVALID_FACILITY)` for invalid enums.

* **`stumpless_get_facility_enum_from_buffer`**:

  * Calls `clear_error()` on success.
  * Calls `raise_error(STUMPLESS_INVALID_FACILITY)` for unmatched strings.

* **`stumpless_get_facility_enum`**:

  * Now clears error if the call to `*_from_buffer` was successful.

---

###  Test Additions (`test/function/facility.cpp`)

Added tests in a new **`FacilityErrorHandling`** suite:

* `ClearsPreviousErrorOnSuccess`: ensures previous errors are cleared.
* `InvalidFacilityEnumRaisesError`: checks invalid enum detection.
* `InvalidFacilityStringRaisesError`: checks invalid string input.

---

###  Function Target

* Added a `function-test-facility` CMake target to run facility tests independently (mirroring `function-test-version` pattern).
* All 8 tests pass successfully.

image

---

###  Doxygen Header Updates (`include/stumpless/facility.h`)

Each relevant function comment now includes:

* `@error This function clears any previous error on success.`
* `@error This function sets STUMPLESS_INVALID_FACILITY on invalid input.`

---

### DIFF
--- a/include/stumpless/facility.h
+++ b/include/stumpless/facility.h
@@ -26,59 +26,59 @@
  */
 
 #ifndef __STUMPLESS_FACILITY_H
-#  define __STUMPLESS_FACILITY_H
+#define __STUMPLESS_FACILITY_H
 
-#  include <stumpless/config.h>
-#  include <stumpless/generator.h>
-#  include <stddef.h>
+#include <stumpless/config.h>
+#include <stumpless/generator.h>
+#include <stddef.h>
 
-#  ifdef STUMPLESS_SYSLOG_H_COMPATIBLE
-#    include <syslog.h>
-#  endif
+#ifdef STUMPLESS_SYSLOG_H_COMPATIBLE
+#include <syslog.h>
+#endif
 
 /**
  * Kernel message facility code value.
  *
  * @since release v2.0.0.
  */
-#  ifdef STUMPLESS_SYSLOG_H_COMPATIBLE
-#    define STUMPLESS_FACILITY_KERN_VALUE LOG_KERN
-#  else
-#    define STUMPLESS_FACILITY_KERN_VALUE 0
-#  endif
+#ifdef STUMPLESS_SYSLOG_H_COMPATIBLE
+#define STUMPLESS_FACILITY_KERN_VALUE LOG_KERN
+#else
+#define STUMPLESS_FACILITY_KERN_VALUE 0
+#endif
 
 /**
  * User-level message facility code value.
  *
  * @since release v2.0.0.
  */
-#  ifdef STUMPLESS_SYSLOG_H_COMPATIBLE
-#    define STUMPLESS_FACILITY_USER_VALUE LOG_USER
-#  else
-#    define STUMPLESS_FACILITY_USER_VALUE ( 1 << 3 )
-#  endif
+#ifdef STUMPLESS_SYSLOG_H_COMPATIBLE
+#define STUMPLESS_FACILITY_USER_VALUE LOG_USER
+#else
+#define STUMPLESS_FACILITY_USER_VALUE (1 << 3)
+#endif
 
 /**
  * Mail system facility code value.
  *
  * @since release v2.0.0.
  */
-#  ifdef STUMPLESS_SYSLOG_H_COMPATIBLE
-#    define STUMPLESS_FACILITY_MAIL_VALUE LOG_MAIL
-#  else
-#    define STUMPLESS_FACILITY_MAIL_VALUE ( 2 << 3 )
-#  endif
+#ifdef STUMPLESS_SYSLOG_H_COMPATIBLE
+#define STUMPLESS_FACILITY_MAIL_VALUE LOG_MAIL
+#else
+#define STUMPLESS_FACILITY_MAIL_VALUE (2 << 3)
+#endif
 
 /**
  * System daemons facility code value.
  *
  * @since release v2.0.0.
  */
-#  ifdef STUMPLESS_SYSLOG_H_COMPATIBLE
-#    define STUMPLESS_FACILITY_DAEMON_VALUE LOG_DAEMON
-#  else
-#    define STUMPLESS_FACILITY_DAEMON_VALUE ( 3 << 3 )
-#  endif
+#ifdef STUMPLESS_SYSLOG_H_COMPATIBLE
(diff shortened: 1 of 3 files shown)


goatshriek:
Thanks very much for putting this together, particularly for describing your updates so thoroughly in the PR!

As you work through these errors (the header check static analysis should help), please also remove the formatting changes from this PR. The changes should be limited to just the changes you are adding, and not reformatting if other parts of updated files.

rs0125:
@goatshriek would it be possible to run a check now? 


There was a missing header somewhere which was causing the issues (private/error.h) 
Added that and that should have fixed it. 
Also cleaned up all the reformatting changes.

Also is there any way I can run the CI tests on my own (like during some other contribution) before sending the commit?

goatshriek:
> Also is there any way I can run the CI tests on my own (like during some other contribution) before sending the commit?

There are two straightforward ways to do this, the first and probably easiest being to enable GitHub Actions on your fork. That should allow you to recreate the exact checks being run for the PR without waiting for a maintainer to approve the runs. You can also go through the workflow files and run the tests manually yourself in your dev environment, but this requires you to install dependencies and tooling, and is a bit more work and more error prone.

rs0125:
Had another header file issue

This should fix it

include/stumpless/facility.h
```diff
@@ -389,6 +389,9 @@ stumpless_get_facility_string( enum stumpless_facility facility );
  * @since release v2.1.0.
  *
  * @param facility_string The facility name to get the enum from.
+ * 
+ * @error Clears previous error on success.
+ * @error Raises STUMPLESS_INVALID_FACILITY on invalid input.
```
goatshriek: As far as I can tell from the [Doxygen documentation](https://www.doxygen.nl/manual/commands.html) `@error` is not a supported special command. Instead, you can just put these two descriptions into their own paragraph to describe the error behaviors.

include/stumpless/facility.h
```diff
@@ -415,6 +418,9 @@ stumpless_get_facility_enum( const char *facility_string );
  * @param facility_string The facility name to get the enum from.
  * 
  * @param facility_buffer_length The length of the buffer
+ * 
+ * @error Clears previous error on success.
+ * @error Raises STUMPLESS_INVALID_FACILITY on invalid input.
```
goatshriek: Same here as above, replace the non-existent `@error` commands with a description of the behavior.

src/facility.c
```diff
@@ -29,44 +31,58 @@ static char *facility_enum_to_string[] = {
...
- return -1;
+  if( result != -1 ) {
+    clear_error();
+  }
+
+  return result;
 }
 
+
+
```
goatshriek: Remove these extra newlines that were added.

test/function/facility.cpp
```diff
@@ -116,4 +116,26 @@ namespace {
     EXPECT_EQ( result, -1 );
   }
 
+  TEST( FacilityErrorHandling, ClearsPreviousErrorOnSuccess ) {
```
goatshriek, line 120: Please indent the body of each new test function.

goatshriek (changes requested):
Overall this looks good! Just a few formatting changes for you to make, and this should be good to go once the headers are also worked out.

rs0125:
That should be all. Please let me know if there's any other changes
Thanks a lot for your patience

rs0125:
I wonder why the test scripts need reimports of headers as different instances when they're already calling scripts from the base facility.c file. The CI workflow keeps flagging it even though it runs on compilation. Is there any particular reason for having that check?

Anyway I've ammended my last commit. Pardon me for making this go on for longer than it should have.

goatshriek (approved):

goatshriek:
Thanks again for sticking with this through the change requests!

The tests are checked separately for includes since they are different compilation units, and likely even use a different compiler than the .c files in the library itself (gcc vs. g++, for example). It's important to make sure that everything used in a single *.cpp or *.c file is included in it, so that there is not an implicit dependency on one header including another, which could lead to unexpected breakage. [IWYU](https://include-what-you-use.org/) is a standard tool for enforcing this, and has more information on the philosophy, if you'd like to learn more about it. We use a custom script here for the checks, to reduce the number of dependencies used for dev, since Ruby is already used for C++ binding generation.

rs0125:
Thanks a lot for the thorough reviews and guidance throughout! Looking forward to contributing more to this project in the future. 

Really appreciate the time and feedback

PR #506, opened 2025-06-13
feat: added test for non cache owned memory fallback, refactored cache_free

### LINKED ISSUE #441: test cache free of non-cache memory
Stumpless uses a simple slab cache implementation in `src/cache.c` to consolidate memory allocations for common structures like entries. [Test coverage](https://app.codecov.io/gh/goatshriek/stumpless/blob/latest/src%2Fcache.c) of this cache is fairly complete, but still has some gaps.

One such gap is what happens when free is called on an object that is not owned by the cache. The mutex for the cache should be unlocked once this routine is finished, but currently this is an untested behavior. You will add a test for this case.

## General Approach

First, familiarize yourself with the `cache_free` implementation in `src/cache.c`. The last line of this function after the for loop is currently uncovered by tests, and is only hit when the entire cache is searched but the object that has been freed is not found.

Next, devise a test for the freeing of an entry structure that is not in the cache. One way to do this is to add a test to `test/function/entry.cpp` that constructs a new entry normally, allocates a new entry struct dynamically and uses `memcpy` to copy the normal entry, and then calls the entry destructor on the manually allocated one. Remember that you will need to manually free the dynamic structure memory as well, since the `cache_free` call will not. Look at the other destructor tests and the `create_empty_entry` helper function for inspiration for this test.

Make sure that you have full coverage on `cache_free` after this test. See the [test documentation](https://github.com/goatshriek/stumpless/blob/latest/docs/test.md#test-coverage) for details on checking coverage.

Finally, refactor `cache_free` after this to simplify the function to having only one exit pointer. This should be as simple as changing the unlock and return inside the for loop to a break statement. Do this _after_ you have written your test, so that you are sure that the refactored code still works!

rs0125:
Fixes #441 

This PR adds a test to cover the fallback behavior in `cache_free()` when the object being freed is **not found in any cache page**. This triggers the final `config_unlock_mutex()` call outside the loop — a line previously uncovered.

To support this, a small refactor was also made to simplify the control flow in `cache_free` by replacing `return` with `break` and allowing a single exit path at the end of the function:

```c
      break;
    }
  }

  config_unlock_mutex( &c->mutex );
}
```

---

### What the Test Does

```cpp
TEST( EntryDestructTest, FreeUncachedEntry ) {
  struct stumpless_entry *original = create_empty_entry();
  ASSERT_NE(original, nullptr);

  struct stumpless_entry *manual = (struct stumpless_entry *) malloc(sizeof(*manual));
  ASSERT_NE(manual, nullptr);
  memcpy(manual, original, sizeof(*manual));

  stumpless_destroy_entry_only(original);
  stumpless_destroy_entry_only(manual);
  free(manual);
}
```

This creates a valid `stumpless_entry`, then copies it into a manually allocated one (`manual`). Since the second one isn’t managed by the slab cache, it forces `cache_free` to run through the loop and hit the unlock path.

---

### Valgrind Warning: Double Free

This version causes a double free warning under Valgrind because `manual` and `original` point to the same internal fields (`message`, `app_name`, `msgid`, etc.). Both get freed in `stumpless_destroy_entry_only()`.

The test still executes correctly and reliably triggers the intended fallback behavior, despite this shared-memory issue, so any guidance to fix that would be appreciated.

---

### Alternate Approaches Explored

* **Direct `cache_free()` call**: Tried calling it via extern + entry_cache, but this isn’t part of the public API. Compiling this from C++ caused issues with _Atomic due to stdatomic.h
* **Zeroed-out struct**: Allocated a malloc'd block and zeroed it out with memset, but calling stumpless_destroy_entry_only() on it triggered pthread_mutex_destroy() on uninitialized memory → segfault.
* **Deep copy of internals**: Too complex and fragile for a unit test — would require safe string duplication and mutex init logic, a lot of which isn't exposed to the public API (for example: the mutex functions not exposed, leading to the necessity of using direct thread interfacing using libraries like pthread.h in the test file)

The current test is a practical compromise to verify the fallback path in `cache_free()` is exercised without rewriting internals or bypassing the public API.

### DIFF
--- a/test/function/entry.cpp
+++ b/test/function/entry.cpp
@@ -2995,4 +2995,17 @@ namespace {
     stumpless_destroy_entry_and_contents(large_entry);
   }
 
+  TEST( EntryDestructTest, FreeUncachedEntry ) {
+    struct stumpless_entry *original = create_empty_entry();
+    ASSERT_NE(original, nullptr);
+
+    struct stumpless_entry *manual = (struct stumpless_entry *) malloc(sizeof(*manual));
+    ASSERT_NE(manual, nullptr);
+    memcpy(manual, original, sizeof(*manual));
+
+    stumpless_destroy_entry_only(original);
+    stumpless_destroy_entry_only(manual);
+    free(manual);
+  }
+
 }
--- a/test/function/entry.cpp
+++ b/test/function/entry.cpp
@@ -2995,17 +2995,16 @@ namespace {
     stumpless_destroy_entry_and_contents(large_entry);
   }
 
-  TEST( EntryDestructTest, FreeUncachedEntry ) {
-    struct stumpless_entry *original = create_empty_entry();
-    ASSERT_NE(original, nullptr);
+TEST( EntryDestructTest, FreeUncachedEntry ) {
 
-    struct stumpless_entry *manual = (struct stumpless_entry *) malloc(sizeof(*manual));
-    ASSERT_NE(manual, nullptr);
-    memcpy(manual, original, sizeof(*manual));
+  struct stumpless_entry *original = create_empty_entry();
+  ASSERT_NE(original, nullptr);
+  
+  struct stumpless_entry manual;
 
-    stumpless_destroy_entry_only(original);
-    stumpless_destroy_entry_only(manual);
-    free(manual);
-  }
+  memcpy(&manual, original, sizeof(manual));
+
+  stumpless_destroy_entry_only(&manual);
+}
 
 }
--- a/src/cache.c
+++ b/src/cache.c
@@ -169,8 +169,7 @@ cache_free( const struct cache *c, const void *entry ) {
       entry_index = ( ( const char * ) entry - current_page ) / c->entry_size;
       locks[entry_index] = 0;
 
-      config_unlock_mutex( &c->mutex );
-      return;
+      break;
     }
   }
 


test/function/entry.cpp
```diff
@@ -2995,4 +2995,17 @@ namespace {
     stumpless_destroy_entry_and_contents(large_entry);
   }
 
+  TEST( EntryDestructTest, FreeUncachedEntry ) {
+    struct stumpless_entry *original = create_empty_entry();
+    ASSERT_NE(original, nullptr);
+
+    struct stumpless_entry *manual = (struct stumpless_entry *) malloc(sizeof(*manual));
```
goatshriek: Just put this structure on the stack, it doesn't need to be dynamically allocated.

test/function/entry.cpp
```diff
@@ -2995,4 +2995,17 @@ namespace {
...
+    ASSERT_NE(manual, nullptr);
+    memcpy(manual, original, sizeof(*manual));
+
+    stumpless_destroy_entry_only(original);
```
goatshriek: Don't bother destroying the original, just do the call on the manual copy.

test/function/entry.cpp
```diff
@@ -2995,4 +2995,17 @@ namespace {
...
+    stumpless_destroy_entry_only(manual);
+    free(manual);
```
goatshriek: Putting the structure on the stack avoids the need for this free call as well.

goatshriek (changes requested):
I made some comments suggesting an alternative setup, give it a try and see if it works.

rs0125:
Just did the changes and the test now runs!

image

The LCOV Coverage also showed that the required mutex unlock line is hit

image


Thanks a lot for the guidance @goatshriek 
Kindly let me know if there's any other change!

goatshriek (approved):

rs0125:
Hi @goatshriek — just a quick reminder on this one whenever you get a chance!
Thanks again for the review and help 🙌

goatshriek:
Sorry for the delay, I needed to find some time to sit down and go through this for the merge.

Thanks again for the contribution, and for following through on the coverage and refactoring! The detailed descriptions and quick responses are especially appreciated.

rs0125:
No worries at all — thanks so much for taking the time to review and merge it!
Really enjoyed contributing, and I appreciate the helpful feedback throughout

PR #509, opened 2025-07-06
documenting private element functions

lucasaponso:
Hello, this is my first PR for the stumpless project, I have noticed that the element functions require documentation. I have provided function headers. Please review. Thanks.

#371 

[Guidelines for Contributing](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md).

### DIFF
--- a/include/private/element.h
+++ b/include/private/element.h
@@ -37,9 +37,27 @@ for( i = 0; i < ( ELEMENT )->param_count; i++ ) {   \
     continue;                                       \
   }
 
+/**
+ * @brief The following function locks the mutex associated
+ * with the current element.
+ * 
+ * Note: The config_lock_mutex locks the elements mutex,
+ * under the hood, pthread_lock is used
+ *
+ * @param element ptr to a stumpless_element
+ * @return void
+ */
 void
 lock_element( const struct stumpless_element *element );
 
+/**
+ * @brief The following function looks up a parameter in an
+ * array.
+ * @param[in] element - A ptr to a element
+ * @param[in] index - The index to look up
+ * @return Pointer to the parameter at the specified index, or NULL if the
+ *         index is out of bounds.
+ */
 struct stumpless_param *
 locked_get_param_by_index( const struct stumpless_element *element,
                            size_t index );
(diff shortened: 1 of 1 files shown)


include/private/element.h
```diff
@@ -37,9 +37,27 @@ for( i = 0; i < ( ELEMENT )->param_count; i++ ) {   \
     continue;                                       \
   }
 
+/**
+ * @brief The following function locks the mutex associated
+ * with the current element.
+ * 
+ * Note: The config_lock_mutex locks the elements mutex,
+ * under the hood, pthread_lock is used
```
goatshriek: `pthread_lock` is only used in builds where pthreads is the chosen multithreading implementation. For example, Windows builds use `EnterCriticalSection` instead. You can just leave this note out - the brief description already describes the functionality well enough.

include/private/element.h
```diff
@@ -37,9 +37,27 @@ for( i = 0; i < ( ELEMENT )->param_count; i++ ) {   \
...
+ *
+ * @param element ptr to a stumpless_element
+ * @return void
```
goatshriek: You can leave out `void` return descriptions.

goatshriek (changes requested):
Thanks for putting together this update! There are a few small updates for you to make below.

For all of these, you also need to add the POSIX safety attribute documentation (thread safety, async safety, and async cancel safety). The link in the issue describes what each of these are, and you can look at other documentation in the library headers to see how this is done in this project.

include/private/element.h
lucasaponso: Ok great, I'll make sure to change that.

include/private/element.h
lucasaponso: Sure.

include/private/element.h
```diff
@@ -37,9 +37,51 @@ for( i = 0; i < ( ELEMENT )->param_count; i++ ) {   \
     continue;                                       \
   }
 
+/**
+ * @brief The following function locks the mutex associated
+ * with the current element.
+ * 
+ * **Thread Safety: MT-Unsafe**
+ * This function is not thread safe as it destroys resources that other threads
+ * would use if they tried to reference this struct.
```
goatshriek, line 46: You'll need to update these safety attributes for each of the functions (not just this one). For example, the `lock_element` function is thread-safe, as it is intended to be used to guarantee thread safety. Be sure you update the reasoning for this as well, for example while `lock_element` is AS-Unsafe, it doesn't have a heap component to that as it doesn't deal with memory; only a lock component.

goatshriek (changes requested):
You'll need to work through the logic and update the POSIX safety attributes for each function. It looks like the current attributes are copy/pasted from another function, which is fine as a starting point but will require updating.

goatshriek:
Checking in to see if you plan to continue work on this, per the [inactivity policy](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md#you-still-there).

goatshriek:
Closing as stale.

PR #517, opened 2025-08-24
Docs/private funcs 371

ritesh006:
Doxyzen style docs added for private function

### DIFF
--- a/include/private/element.h
+++ b/include/private/element.h
@@ -25,7 +25,23 @@
 #  include <stumpless/param.h>
 #  include "private/param.h"
 
-#  define FOR_EACH_PARAM_WITH_NAME( ELEMENT, NAME ) \
+/**
+ * Iterates the parameters of an element and executes the following block only
+ * for parameters whose name matches the provided name.
+ *
+ * **Thread Safety: MT-Safe req:element-stable**
+ * This macro locks each parameter during the name comparison but does not guard
+ * against concurrent structural changes to the element's parameter array. No
+ * other thread should modify the element's parameter list during iteration.
+ *
+ * **Async Signal Safety: AS-Unsafe lock**
+ * This macro is not safe to use from signal handlers due to locking.
+ *
+ * **Async Cancel Safety: AC-Unsafe lock**
+ * This macro is not safe for threads that may be asynchronously cancelled
+ * during the locking operations performed inside the macro.
+ */
+#define FOR_EACH_PARAM_WITH_NAME( ELEMENT, NAME ) \
 for( i = 0; i < ( ELEMENT )->param_count; i++ ) {   \
   param = element->params[i];                       \
                                                     \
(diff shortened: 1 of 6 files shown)


include/private/element.h
```diff
@@ -25,7 +25,23 @@
 #  include <stumpless/param.h>
 #  include "private/param.h"
 
-#  define FOR_EACH_PARAM_WITH_NAME( ELEMENT, NAME ) \
+/**
+ * Iterates the parameters of an element and executes the following block only
+ * for parameters whose name matches the provided name.
+ *
+ * **Thread Safety: MT-Safe req:element-stable**
```
goatshriek: I can't find anything regarding the `req` safety attribute remark in the attributes man page. Where did you get this remark from?

include/private/element.h
```diff
@@ -37,9 +53,49 @@ for( i = 0; i < ( ELEMENT )->param_count; i++ ) {   \
...
 
+/**
+ * Gets a parameter by index from a locked element, without performing additional locking.
+ *
+ * **Thread Safety: MT-Safe req:element-locked**
+ * This function is safe when the caller already holds the element's lock. It
+ * must not be called without the element being locked by the caller.
+ *
+ * **Async Signal Safety: AS-Unsafe**
+ * This function is not intended for use in signal handlers.
```
goatshriek: Provide a reason that this is AS-Unsafe, particularly since it is marked as AC-Safe.

goatshriek (changes requested):
Thanks for putting this together! Just a few changes/questions for you to address.

ritesh006:
Addressed feedback:

Replace req: with MT-Safe race:element; note no concurrent changes to element->params/param_count.

Macro: AS-Unsafe (lock+strcmp), AC-Unsafe (lock).

locked_get_param_by_index: AS-Safe / AC-Safe + caller must hold element lock.

Docs only; no behavior change.

ritesh006:
Hi @goatshriek I Have addressed your comment please review this

include/private/element.h
```diff
@@ -25,7 +25,23 @@
...
+/**
+ * Iterates the parameters of an element and executes the following block only
+ * for parameters whose name matches the provided name.
+ *
+ * **Thread Safety: MT-Safe race:element**
+ * Requires external synchronization so that element->params and
+ * element->param_count are not modified during iteration. Each param is locked
+ * only for the duration of the strcmp() comparison.
+ *
+ * **Async Signal Safety: AS-Unsafe lock strcmp**
```
goatshriek: `strcmp` is not a recognized unsafe attribute (see [this list](https://www.gnu.org/software/libc/manual/html_node/Unsafe-Features.html)). Remove it from the header and leave it in the description below.

goatshriek (changes requested):
Just one more chance and this should be ready to go.

include/private/element.h
```diff
@@ -25,7 +25,23 @@
...
+ * element->param_count are not modified during iteration. Each param is locked
+ * only for the duration of the strcmp() comparison.
+ *
+ * **Async Signal Safety: AS-Unsafe lock**
+ * Uses lock_param/unlock_param (mutex operations) and strcmp(), which are not
+ * async-signal-safe.
+ *
+ * **Async Cancel Safety: AC-Unsafe lock**
+ * May block at mutex operations, which are cancellation points.
+*/
```
goatshriek: Correct this indentation error by adding a space before the `*/`, which is likely what is causing the failure in the single file build (due to the way it parses the headers).

goatshriek (changes requested):
One more minor change before this can be accepted, due to some failures the CI runs noted.

goatshriek (approved):

ritesh006:
here is something due to branch is not merged.

goatshriek:
Thanks for your contribution and patience in working through the requested changes!

PR #518, opened 2025-08-26
Add Polish translation

Maqi-x:
## Add Polish translation
This PR adds the initial Polish translation for the project.


goatshriek:
Thanks for putting this contribution together! I believe the failure is unrelated to your changes, but I'll need to get that worked out before I pull this in.

goatshriek (changes requested):
The static analysis CI run noted two strings that are longer than 80 characters - once those are corrected this should be ready to accept.

Maqi-x:
Thanks for pointing that out! I've fixed the long lines, should be good now.

goatshriek (approved):

goatshriek:
Thanks for your contribution and patience!

PR #521, opened 2025-08-29
docs(private): add Doxygen for error.h

ritesh006:
Doxygen Doc Added for error.h.

### DIFF
--- a/include/private/error.h
+++ b/include/private/error.h
@@ -23,13 +23,47 @@
 #include <stumpless/error.h>
 #include "private/config.h"
 
+/**
+ * Clears the current thread's error state.
+ *
+ * **Thread Safety: MT-Safe**
+ * This function is thread safe.
+ *
+ * **Async Signal Safety: AS-Safe**
+ * Modifies only thread-local error state; no locks, allocation, or I/O.
+ *
+ * **Async Cancel Safety: AC-Safe**
+ * Contains no cancellation points.
+ */
 void
 clear_error( void );
 
+/**
+ * Raises an error indicating that a network address operation failed.
+ *
+ * @param message   Localized description of the failure.
+ * 
+ * @param code      Error code produced by the failing call.
+ * 
+ * @param code_type Localized description of what @p code represents.
+ *
+ * **Thread Safety: MT-Safe**
+ * **Async Signal Safety: AS-Unsafe**
+ * **Async Cancel Safety: AC-Unsafe**
+ */
 COLD_FUNCTION
 void
 raise_address_failure( const char *message, int code, const char *code_type );
 
+/**
+ * Raises an error indicating that a required argument was empty.
+ *
+ * @param message Localized message describing the error.
+ *
+ * **Thread Safety: MT-Safe**
+ * **Async Signal Safety: AS-Unsafe**
+ * **Async Cancel Safety: AC-Unsafe**
+ */
 COLD_FUNCTION
 void
 raise_argument_empty( const char *message );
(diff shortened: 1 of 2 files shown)


goatshriek (changes requested):
Please do a rebase on this so the previous changes to element.h are not included.

What is the rationale behind `clear_error` being AS-Unsafe and AC-Unsafe? I am not seeing an execution path where either of these execution scenarios cause a problem.

ritesh006:
Rebased the branch onto latest main so only error.h changes remain.

goatshriek (changes requested):
`element.h` is still included in this change.

ritesh006:
element.h removed from this commit

include/private/error.h
```diff
@@ -99,14 +163,35 @@ raise_error( enum stumpless_error_id id,
...
+ * **Thread Safety: MT-Safe**
+ * **Async Signal Safety: AS-Unsafe**
+ * **Async Cancel Safety: AC-Unsafe**
+ */
 COLD_FUNCTION
 void
 raise_file_write_failure( void );
 
+/**
+ * Raises an error indicating that a file write failed.
```
goatshriek: Copy/paste error here - update to describe the actual function.

include/private/error.h
```diff
@@ -241,22 +415,74 @@ COLD_FUNCTION
...
+ * **Async Cancel Safety: AC-Unsafe**
+ */
 COLD_FUNCTION
 void
 raise_socket_connect_failure( const char *message,
                               int code,
                               const char *code_type );
 
+/**
+ * Raises an error indicating a generic socket failure.
```
goatshriek: This is not a generic failure function, it is for when the `socket` function has failed.

goatshriek (changes requested):
A few missed details below, and then this will be ready to merge.

ritesh006:
comment is addressed please take a look

include/private/error.h
```diff
@@ -99,24 +163,74 @@ raise_error( enum stumpless_error_id id,
...
+ * **Thread Safety: MT-Safe**
+ * **Async Signal Safety: AS-Unsafe**
+ * **Async Cancel Safety: AC-Unsafe**
+ */
 COLD_FUNCTION
 void
 raise_file_write_failure( void );
 
+/**
+ * Raises an error indicating a function-target operation failed.
```
goatshriek: This description doesn't provide any information beyond what the name of the function already does, but in this case there isn't enough in the name to make it clear. For example, if I search for "function-target" in the code base the hits don't help me understand what this is.

You need to describe what this function is actually used to represent. Start by searching for the function name, and understanding what actually causes this error to be raised.

goatshriek (changes requested):
Another clarification in this function description, since the description as is does not provide any additional information.

include/private/error.h
ritesh006: Thanks for the guidance! I traced the call sites and clarified the docs.

What it represents: this error is raised in the send_entry_to_function_target(...) path when a function target delivers an entry via an app-provided callback (log_function).

Exact trigger: if log_function(target, entry) returns a negative value, we call raise_function_target_failure(result).

What code means: it’s the callback’s return value (the negative status from log_function) forwarded for diagnosis.

I updated the raise_function_target_failure(int) comment to spell out the above (no behavior change). If you prefer a different term than “function target” (e.g., “callback target”), I can switch wording in the doc as well. Thanks again!

ritesh006: Thanks for the guidance! I traced the call sites and clarified the docs.

What it represents: this error is raised in the send_entry_to_function_target(...) path when a function target delivers an entry via an app-provided callback (log_function).

Exact trigger: if log_function(target, entry) returns a negative value, we call raise_function_target_failure(result).

What code means: it’s the callback’s return value (the negative status from log_function) forwarded for diagnosis.

I updated the raise_function_target_failure(int) comment to spell out the above (no behavior change). If you prefer a different term than “function target” (e.g., “callback target”), I can switch wording in the doc as well. Thanks again!

PR #525, opened 2025-09-10
tests(function): inthelper.c code coverage

ritesh006:
Test code coverage for inthelper


goatshriek (changes requested):
The test suite does not currently support direct calling of private functions, so you won't be able to simply call functions from `inthelper.h`. You'll need to find somewhere this is used and influence through public functions.

Also, please review other tests, and match yours with theirs. This should be a C++ test that uses GTest, not a custom executable. See [docs/test.md](https://github.com/goatshriek/stumpless/blob/latest/docs/test.md) for details.

goatshriek (changes requested):
This test does not add any additional coverage, and it isn't clear to me how you expected it to? Please ensure that coverage is in fact increased with local testing before nominating this for review. You can find information on testing for coverage in the test documentation linked previously.

Please explain why the `runtime_size` hack is necessary here?

ritesh006:
image

test/function/inthelper.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+namespace {
+
+class InthelperPublicTest : public ::testing::Test {
+protected:
+  stumpless_element* element = nullptr;
+
+  void SetUp() override {
+    element = stumpless_new_element("e");
+    ASSERT_NOT_NULL(element);
```
goatshriek, line 17: assertions should not be made in setup or teardown code

test/function/inthelper.cpp
```diff
@@ -0,0 +1,51 @@
...
+    ASSERT_NOT_NULL(p);
+    const stumpless_element* add_res = stumpless_add_param(element, p);
+    ASSERT_EQ(add_res, element);
+    EXPECT_NO_ERROR;
+  }
+
+  void TearDown() override {
+    if (element) {
+      stumpless_destroy_element_and_contents(element);
+      element = nullptr;
```
goatshriek, line 28: this clearing of the variable is unnecessary as the instance is being torn down

test/function/inthelper.cpp
```diff
@@ -0,0 +1,51 @@
...
+    }
+    stumpless_free_all();
+  }
+};
+
+TEST_F(InthelperPublicTest, GetParamByIndex_WithinRange) {
```
goatshriek, line 34: This test is duplicative of existing tests and does not need to be included.

test/function/inthelper.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+TEST_F(InthelperPublicTest, GetParamByIndex_TooLargeIndex) {
+#if SIZE_MAX > INT_MAX
+  size_t huge = static_cast<size_t>(INT_MAX) + 1u;
+#else
+  size_t huge = static_cast<size_t>(INT_MAX); // platform fallback
+#endif
+  const stumpless_param* param = stumpless_get_param_by_index(element, huge);
+  EXPECT_NULL(param);
+  EXPECT_ERROR_ID_EQ(STUMPLESS_INDEX_OUT_OF_BOUNDS);
```
goatshriek, line 48: This test does improve the `size_t_to_int` coverage, but belongs in the existing `test/function/element.cpp` test suite alongside the other tests of `stumpless_get_param_by_index`. Also, it should check the code of the error to ensure that it matches with what the wraparound functionality is (`-1` in this case).

test/function/inthelper.cpp
```diff
@@ -0,0 +1,51 @@
...
+};
+
+TEST_F(InthelperPublicTest, GetParamByIndex_WithinRange) {
+  const stumpless_param* param = stumpless_get_param_by_index(element, 0u);
+  ASSERT_NOT_NULL(param);
+  EXPECT_NO_ERROR;
+}
+
+TEST_F(InthelperPublicTest, GetParamByIndex_TooLargeIndex) {
+#if SIZE_MAX > INT_MAX
```
goatshriek, line 41: This conditional does have some value in making sure the test is relevant, but should only be wrapped around the validation code of the error code mentioned below. It doesn't need to be used to modify the argument to the function here.

goatshriek (changes requested):
The `TEST_F(InthelperPublicTest, GetParamByIndex_TooLargeIndex)` test is the only one that you need to keep here, and I believe you should be able to move it to the existing `test/function/element.cpp` test suite. It fits better with other tests of the `stumpless_get_param_by_index` function, which are located there. You should also rename it to fit with the other tests of that function in the existing test suite. There are also some tuning notes on this test included in the review below for you to take into account.

ritesh006:
this going quite hard closing this pr for now and would like to open new pr

Issue #478, opened 2024-11-09
stumpless_set_entry_prival allows invalid privals


goatshriek:
As discussed in #476. the prival is not appropriately validated in this function, allowing values greater than 256 through.

alexdeluera:
I would like to work on this issue!

FujiwaraNaoto:
I have added validation.

goatshriek:
Resolved by #526.

PR #526, opened 2025-09-11
Add prival validation to stumpless_set_entry_prival() (modification for #478 issue)

FujiwaraNaoto:
I added validation for `prival` in the `stumpless_set_entry_prival()` function. which is mentioned [here](https://github.com/goatshriek/stumpless/issues/478)

This is my very first OSS commit merge request. I would appreciate your review.


goatshriek (changes requested):
Overall this looks great, thank you for putting this together!

It looks like there is a missing locale header update, which caused the locale tests to fail.

The memory leak failures can likely be resolved by calling `stumpless_destroy_entry_only` on the entries created in the tests before `stumpless_free_all`.

FujiwaraNaoto:
Thanks for your reply and concrete advice.
I have modified codes.
Could you review it again?

goatshriek:
All that remains is one more line that is too long, and some re-ordering for alphabetic ordering within the locale headers.

FujiwaraNaoto:
Thanks for your reply.
I have modified the codes , run `ruby scripts/check_l10n.rb "include/private/config/locale/*-??.h"`.
and nothing is displayed, so tests will be passed.

I have to make some changes

FujiwaraNaoto:
> I have to make some changes

modified.

goatshriek (approved):
Excellent work, thank you for sticking with this!

FujiwaraNaoto:
Thanks for merging my merge request.

By the way I don't understand why the results of codecov/project change before and after the commit [acb2061](https://github.com/goatshriek/stumpless/pull/526/commits/acb2061e72b64ea0c6656f2894b712be848d07ad)

goatshriek:
The codecov testing is based on the CI runs generating coverage reports, which include several different configurations. These all generate different coverage, which is good! That's why we have them.

Unfortunately, it can also mean that a coverage report that was generated halfway through the CI runs makes it look like coverage dropped when it didn't. If you're watching coverage before all of the runs are completed, you may see a report that makes it looked like it is lower, when in reality it just isn't done yet.

Some of the thread-safety tests are also non-deteriministic because of the nature of concurrency, which generates some false changes in coverage that aren't your fault. In this case that may be what's happening, since I see some journald test coverage changes, which is one of the test cases that is variable right now. That's obviously not the desired end state, I just haven't resolved it yet since it isn't critical.

All that to say, don't worry about any minor coverage changes for contributions! What's important is the coverage of your patch, which is covered 100%, and so you are good to go.

PR #535, opened 2025-10-15
Use library strncasecmp if available

### LINKED ISSUE #531: use library strncasecmp when available
As discussed in #437, the implementation of `strncasecmp` is currently done manually to provide portability on systems that do not have it, such as Windows. The desired end state is to have a configuration step detect whether it is available and only fall back to the custom implementation if it is not.

flinkflonk:
Introduces new flags regarding the availability of strncasecmp (which nowadays should be in strings.h), moves the custom function into its own sourcefile which is only compiled and connected to the rest of the code via a wrapper header.

Closes #531


flinkflonk:
There's one "FIXME" in my code, I haven't checked out the thread safety of the custom function (which I didn't write, I just refactored :grin: ). Also, I have tested this on Linux, but not on Windows, so somebody might want to check if it still compiles on Windows (it should).

goatshriek:
My apologies for the delay, I am aiming to review this and provide comments on the next 24 hours.

include/private/config/have_strncasecmp.h
goatshriek, line 1: I think you can remove this header, and simply include strings.h to get `strncasecmp` in the wrapper header. The BSD case you mentioned would probably get a different, build system-based solution when encountered.

include/private/config/wrapper/strncasecmp.h
```diff
@@ -0,0 +1,36 @@
...
+
+#ifndef __STUMPLESS_PRIVATE_CONFIG_WRAPPER_STRNCASECMP_H
+#define __STUMPLESS_PRIVATE_CONFIG_WRAPPER_STRNCASECMP_H
+
+#include "private/config.h"
+
+/* definition of config_get_now */
+#ifdef HAVE_STRNCASECMP
+#  include "private/config/have_strncasecmp.h"
+#  define strncasecmp_custom strncasecmp
```
goatshriek: I know it will be a little more work to update the usages of this function, but the new wrapper should be named `config_strncasecmp` in keeping with the [portability conventions](https://github.com/goatshriek/stumpless/blob/latest/docs/portability.md) of the project.

src/config/no_strncasecmp.c
```diff
@@ -0,0 +1,35 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <ctype.h>
+#include <stddef.h>
+#include "private/no_strncasecmp.h"
```
goatshriek: I believe you'll need to update this to include the `config` part of the path.

include/private/config/no_strncasecmp.h
```diff
@@ -0,0 +1,54 @@
...
+
+#ifndef __STUMPLESS_PRIVATE_CONFIG_NO_STRNCASECMP_H
+#define __STUMPLESS_PRIVATE_CONFIG_NO_STRNCASECMP_H
+
+#include <stddef.h>
+
+/**
+ * Implements the POSIX 1-2008 function to compare two strings ignoring case
+ *
+ * FIXME: check these:
```
goatshriek: From my understanding, the thread safety is MT-Safe with a `race` attribute that the input strings not be changed during execution.

The async safety depends on the locale due to the use of `tolower`, unfortunately, and so will need a `locale` remark (described [here](https://man7.org/linux/man-pages/man7/attributes.7.html)). I would mark these as unsafe for that reason, and a future effort can introduce a `_r` reentrant version of this if it's needed.

goatshriek (changes requested):
Thanks for putting this together! It looks like there are some failing test cases, which I didn't look to closely at as I expect you'll be able to work out the issues. The other comments should be straightforward to address, and then I believe this will be good to go!

include/private/config/no_strncasecmp.h
flinkflonk: Ok, I take your word for it as I have not much experience with checking out thread- and async-safety. I hope the attributes have been added correctly, feel free to change it if I did something wrong.

flinkflonk:
I have now gone through your code review and made the requested changes. The tests failing are not really related to my code (and they failed before I touched it), and as far as I can see it's because of three factors (which are out of scope for this pull request):
- the function `stumpless_get_severity_enum_from_buffer` in src/severity.c returns a `enum stumpless_severity`, but in two places there's just a `size_t` or `int` returned (`i` or `-1`).
- the test checks for a value of `-1` being returned, but it is clearly a `STUMPLESS_SEVERITY_*_VALUE` in the test cases provided
- besides, `EMERG`, `WARN` and `ERROR` are marked as "deprecated" in the source code for a different test

src/config/no_strncasecmp.c
goatshriek: The Windows build is failing because this header file cannot be found: updating to `#include "private/config/no_strncasecmp.h` should resolve this.

goatshriek (changes requested):
Thanks for digging into that test! The implementation is indeed broken, as you outlined. I'll fix this in a separate change - if you don't mind, please comment out the failing test case, and I will re-enable it when I fix the issue. I'd like to see the rest of the CI tests pass to make sure they're good to go.

Otherwise, I believe this will be good to go once the header path is updated in `src/config/no_strncasecmp.c`, and the header check issues are resolved. The header checks should be resolved by adding entries into the `tools/check_headers/stumpless_private.yml` manifest for the new symbols `HAVE_STRNCASECMP`, `config_strncasecmp` and `strncasecmp_custom`, and an entry into `tools/check_headers/standard_library.yml` for `strncasecmp`.

src/config/no_strncasecmp.c
flinkflonk: Uhm, forgot to save :sheep: - sorry.

flinkflonk:
Ok, after disabling the test the ones running on my fork are complaining about not being able to upload coverage reports. But more important: the Windows build returns success (because it doesn't include coverage), so I guess we're good to go?

goatshriek:
Great, looks good! You don't have to worry about the coverage reports. The only remaining thing is to add the header check entries outlined above, which should resolve the header check that is currently failing in the static analysis.

flinkflonk:
Ok, hopefully I ticked all the boxes and dotted all the i-s, it seems the only thing breaking now - at least in my fork - is "Sonarcloud Analysis" (which can't "find the project", whatever that means).

goatshriek:
Excellent work, thanks once again for putting this change together! I especially appreciate that you checked the CI runs in your fork while working through the fixes - this is a huge time saver and is often missed by contributors.

flinkflonk:
Thanks a lot for being so patient with me, I might be somewhat useful at C (and maybe a little bit of C++) but not really into pull requests and what is expected. Looking at the CI results in my own fork was no biggie here, I've seen projects where it is more of an issue to understand what is going on (openssl for instance has a CI run that takes multiple *hours*).
So thanks for the help from your side, and hope my work added some value :)

PR #545, opened 2025-11-05
Fix facility case-insensitive match (Fixes #438)

### LINKED ISSUE #438: Implement strcasecmp to avoid creating an uppercase copy
When working on issue #437 I noticed that `stumpless_get_facility_enum_from_buffer`and `stumpless_get_severity_enum_from_buffer` create a copy of the parameter and then convert that copy to uppercase to compare.

As a solution I suggest implementing the `strcasecomp` function which performs a case-insensitive compare. And use this function to directly compare the parameter instead of its copy thus further reducing allocations and improving speed.

Maybe I am forgetting something, let me know!

2400032703kotinijeswanth:
Replaced uppercase buffer conversion with config_strncasecmp() to improve performance and simplicity.

### DIFF
--- a/src/facility.c
+++ b/src/facility.c
@@ -1,5 +1,4 @@
 // SPDX-License-Identifier: Apache-2.0
-
 /*
  * Copyright 2018-2022 Joel E. Anderson
  *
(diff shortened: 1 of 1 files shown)


goatshriek:
This change doesn't actually do anything other than change formatting and remove a header that is necessary for the definition of `size_t`.

PR #549, opened 2025-11-25
docs: updated benchmark.md code snippet to match current CopyElement (fixes #444)

### LINKED ISSUE #444: update benchmark documentation to match code

juliafeldhousen:
Updated the CopyElement example in benchmark.md documentation so it matches the current implementation in test/performance/element.cpp. Hopefully I understand what the issue was asking for properly, let me know if anything needs to be changed!

### DIFF
--- a/docs/benchmark.md
+++ b/docs/benchmark.md
@@ -68,9 +68,6 @@ static void CopyElement(benchmark::State& state){
   const struct stumpless_element *result;
 
   INIT_MEMORY_COUNTER( copy_element );
-  stumpless_set_malloc( copy_element_memory_counter_malloc );
-  stumpless_set_realloc( copy_element_memory_counter_realloc );
-  stumpless_set_free( copy_element_memory_counter_free );
 
   element = stumpless_new_element( "copy-element-perf" );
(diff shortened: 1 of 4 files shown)


goatshriek:
Thanks very much for your patience in waiting for this to get reviewed, and for finally finishing an issue that has given many others trouble!

PR #550, opened 2025-12-03
Improve journald initialization

### LINKED ISSUE #487: improve journald initialization
As described in #79, the initialization of several fields for journald targets could be improved, both by using static field initialization and removing the valgrind exception for the code, which is currently the only exception for the project.

SteveNewson:
The current implementation of the journald target creates a copy of the `fixed_fields` structure for each message. This is inefficient and results in a temporary memory leak that valgrind picks up.

This change switches to a thread local (where available) static struct which elimiates the need to copy the memory or free it upon completion.

This change also removes the valgrind suppressions for this problem.

Fixes: #487


goatshriek (changes requested):
This is excellent work, thanks very much for putting it together! It will be a big relief to no longer have any valgrind exceptions.

Just a few extra cleanup changes to add for this change, and then it will be good to go: please also remove `init_fixed_fields` from `include/private/target/journald.h` and `tools/check_headers/stumpless_private.yml`.

SteveNewson:
It was right there in the dev guide...

goatshriek:
Once again, excellent work and thank you for putting this together!

PR #551, opened 2026-02-04
Add Turkish translation for README

Natronabi:
Hi! I'm a first-year software engineering student from Ankara University. I noticed the Turkish flag in the README but the translation link was missing. I've added the Turkish localization for the README file. Hope this helps the project!


goatshriek (changes requested):
Thanks very much for putting this together! Just one adjustment to make and then this will be accepted; please add a link at the top of the main README to the new translation as well, so that it is easy for folks to find and use your work!

PR #553, opened 2026-02-17
Add Irish (ga-ie) locale support

mklinzing:
This pull request adds Irish (ga_IE) localization support to Stumpless by introducing translated message strings and the corresponding locale configuration. The goal is to expand language coverage while maintaining consistency with existing localization patterns and conventions. This change resolves Issue #117 reported on GitHub.

[Guidelines for Contributing](https://github.com/goatshriek/stumpless/blob/latest/docs/CONTRIBUTING.md).


include/private/config/locale/ga-ie.h
```diff
@@ -0,0 +1,308 @@
...
+ */
+
+/** @file
+ * United States English localization strings.
+ */
+
+ //see more Irish Gaelic computing terminology aid here: https://www3.smo.uhi.ac.uk/oduibhin/tearmai/liosta.htm 
+
+#ifndef __STUMPLESS_PRIVATE_CONFIG_LOCALE_EN_US_H
+#  define __STUMPLESS_PRIVATE_CONFIG_LOCALE_EN_US_H
```
goatshriek, line 26: Update this header guard to use `__STUMPLESS_PRIVATE_CONFIG_LOCALE_GA_IE_H`

goatshriek (changes requested):
Thanks very much for putting this together! Just one update to the header and a few lines that need to be broken up so that they aren't too long (see the static analysis run), and then this should be good to go!

PR #554, opened 2026-03-18
Fix strncasecmp_custom

Lemuren:
- Fixes `strncasecmp_custom` so that its behavior matches that of the `libc` implementation.
- Add explicit checks to `stumpless_get_severity_enum_from_buffer` to disallow partial matches even when using `strncasecmp`. I've reenabled the failing test.

The explicit checks in `src/severity.c` are a bit ugly since we're essentially just forcing `strncasecmp` to behave like `strcasecmp`, but if the user doesn't have one they probably will not have the other ;).

### DIFF
--- a/src/config/no_strncasecmp.c
+++ b/src/config/no_strncasecmp.c
@@ -25,14 +25,9 @@
  */
 int
 strncasecmp_custom( const char *s1, const char *s2, size_t n ) {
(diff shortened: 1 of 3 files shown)


goatshriek:
Thanks again for putting this together! It's much appreciated.

PR #558, opened 2026-04-08
Add tests for inthelper helper functions

chithien14td-beep:
## Summary
Add function tests for the private inthelper utilities.

## Changes
- add tests for `cap_size_t_to_int`
- add tests for `size_t_to_int`
- cover normal values, boundary values, and overflow cases
- register the new function test target in CMake

## Validation
- built `function-test-inthelper`
- ran `./build/function-test-inthelper.exe`
- all 14 tests passed


chithien14td-beep:
Hi, thank you for taking the time to review my PR.

In this PR, I added function tests for the private `inthelper` helpers, specifically `cap_size_t_to_int` and `size_t_to_int`, to cover normal values, boundary cases, and overflow-related cases. I also added the test target to the CMake configuration and updated the changelog. I hoped this would be a small but useful improvement to the project’s test coverage for these low-level helper functions.

I’m also new to open-source contribution, and this is my first PR, so I’m still learning the workflow. I noticed that the PR is blocked because the commits must have verified signatures. I checked the contribution documentation before submitting, but I did not realize this was required, so I’m very sorry for missing it. I would really appreciate your guidance on the correct next step if I need to recreate the commit with a verified signature.

Thank you very much for your time and help.

goatshriek (changes requested):
Thanks very much for putting this together! You don't need to worry about the verified signatures for this one, but you will need to make sure that the currently failing checks can pass.

There aren't currently any tests that directly link in source files from the project, and instead only link in the built library and test it directly. The best way to test the integer helpers is to find where they are used within the library and test that functionality. If you want to create separate tests for individual modules, you will need to manually create the tests, since the existing test helpers like `add_function_test` link in stumpless and its dependencies, which shouldn't be done for an isolated white box unit test.

chithien14td-beep:
> Thanks very much for putting this together! You don't need to worry about the verified signatures for this one, but you will need to make sure that the currently failing checks can pass.
> 
> There aren't currently any tests that directly link in source files from the project, and instead only link in the built library and test it directly. The best way to test the integer helpers is to find where they are used within the library and test that functionality. If you want to create separate tests for individual modules, you will need to manually create the tests, since the existing test helpers like `add_function_test` link in stumpless and its dependencies, which shouldn't be done for an isolated white box unit test.

Thanks, that makes sense. I see now that the existing test setup is meant for black-box testing of the built library rather than isolated source-level unit tests.

I'll take another look at where these integer helpers are used in the library and rework the tests around that functionality so the checks can pass.

chithien14td-beep:
Thanks for the clarification. I removed the isolated inthelper test and related CMake changes, and replaced them with function-level coverage through the public index lookup APIs. I added boundary tests for out-of-bounds indices larger than INT_MAX, which exercise the error path that converts size_t indices into the public error code.

The new huge-index tests pass locally. The remaining local failures I’m seeing are existing SetMessageStrTest.* cases in the entry test executable on my Windows/MSYS2 setup, and they do not appear to be related to this change.

goatshriek (changes requested):
These changes look great! The only thing remaining is to update the header list for `INT_MAX`, which doesn't currently include `climits`. This is in `tools/check_headers/standard_library.yml`: you'll just need convert the entry for `INT_MAX` into an array (as other symbols already are in that file) and add `climits` in addition to `limits.h`. This should resolve the failing static analysis CI check.

chithien14td-beep:
Updated the `INT_MAX` header mapping in `tools/check_headers/standard_library.yml` to include `climits` as well.

PR #559, opened 2026-04-10
Feature/italian l10n

DeltaPhi0:
Wrote the Italian translation for the README, I did not import the INSTALL, docs nor assets folders in this and the links in the files point to the english version, The spacing mught be a bit off at times, not exactly like the main english version, I hope that's not a problem


goatshriek:
Thanks very much for putting this together, translation support is always greatly appreciated!!

PR #561, opened 2026-04-14
Add Traditional Chinese (zh-Hant) localization support

fx-maker:
- Add zh-hant.h locale configuration header
- Update CMake and workflow files for zh-Hant
- Update changelog to include Traditional Chinese
- Update README with zh-Hant language option


fx-maker:
I apologize for the inconvenience — this is my first time contributing to an open source project, and I wasn't aware of the commit signature requirement. I'm sorry for missing that.

Could you please advise if commit signing is required for this PR? I'll be happy to sign the commit or rebase with a signed commit if needed.

Thank you for your understanding and for maintaining this great project!

ChangeLog.md
```diff
@@ -9,6 +9,11 @@ fixes, check out the
 [roadmap](https://github.com/goatshriek/stumpless/blob/master/docs/roadmap.md).
 
 
+## [Unreleased]
```
goatshriek, line 12: This will be released in 3.0.0 - please move this to that entry in the Changelog instead.

goatshriek (changes requested):
Thanks very much for putting this together! You don't need to worry about the commit signing requirement for this PR, though I do recommend you work through setting this up on your own at some point, as it's an important concept to understand.

There is a small change to make below, and also an update that needs to be made to `.github/workflows/locale.yml` in order to support this locale: on line 51, the sed expression needs to modified to support subtags with more than two characters. One possible way to do this is to update it to:

`EXPECTED_LANG=$( echo $LANG | sed 's/\(..\)_\(.*\).UTF-8/\1-\2/g' )`

Please get these changes incorporated and this should be ready to accept!

ChangeLog.md
fx-maker: @goatshriek Thanks so much for the detailed feedback and suggestions!

I really appreciate the clear explanation and the fix for the sed expression — that's very helpful. I understand the changelog needs to be moved to the 3.0.0 section now.

I'll make the changes right away and update the PR shortly.

Thanks again for your patience and guidance!

fx-maker:
@goatshriek I've made the changes you requested:

- Moved the changelog entry to the 3.0.0 section
- Updated the sed regex in `locale.yml` to `\(..\)_\(.*\).UTF-8`
- Adjusted `l10n.cmake` to prioritize zh-Hant before zh-CN and added case-insensitive matching
- Removed the accidentally uploaded `Testing/` directory from the branch

Please let me know if there's anything else that needs to be adjusted. Thanks again for your help!

tools/cmake/l10n.cmake
```diff
@@ -35,6 +35,9 @@ elseif(LOCALE MATCHES "^el[_-]")
 elseif(LOCALE MATCHES "^pt[_-]")
   set(STUMPLESS_LANGUAGE "pt-BR")
   set(USE_LOCALE_PT_BR TRUE)
+elseif(LOCALE MATCHES "^zh-hant$" OR LOCALE MATCHES "^zh-Hant$" OR LOCALE MATCHES "^zh-HANT$")
```
goatshriek: This set of matches is what is causing the locale detection to fail, leading to the failing CI check. The locale that is set is using underscores, but all of these patterns only use a dash. Updating them to use the character class of either an underscore or dash `[_-]` as in other patterns should correct this.

goatshriek (changes requested):
Looks like the CI run for this locale is failing - a small adjustment should correct it.

tools/cmake/l10n.cmake
fx-maker: @goatshriek Thanks so much for the detailed review and guidance!

You're right — I used hyphen by default when testing locally and didn't realize the CI uses underscore. Sorry for the oversight.

I've updated the locale matching to support underscore separator as you suggested. Please take another look when you get a chance.

Thanks again for your patience and help!

tools/cmake/l10n.cmake
```diff
@@ -35,6 +35,9 @@ elseif(LOCALE MATCHES "^el[_-]")
 elseif(LOCALE MATCHES "^pt[_-]")
   set(STUMPLESS_LANGUAGE "pt-BR")
   set(USE_LOCALE_PT_BR TRUE)
+elseif(LOCALE MATCHES "^zh[_-][Hh][Aa][Nn][Tt]$")
```
goatshriek: I think the `$` is still causing issues here, due to the `.UTF-8` suffix.

goatshriek (changes requested):
Still a little adjustment needed here, I believe.

tools/cmake/l10n.cmake
fx-maker: Thanks for catching that! I've removed the trailing `$` to allow `.UTF-8` suffix. Please take another look.

goatshriek:
Thanks once again for putting this together, and sticking with it through the changes! Translation support is always hugely appreciated.

PR #563, opened 2026-04-14
test(strbuilder): add complete unit tests for strbuilder module

pengzhongChen:
Summary
This PR adds comprehensive unit tests for the strbuilder module and fixes a minor memory allocation handling issue in src/entry.c.

Changes
1. Added test/function/strbuilder.cpp with 33 test cases covering strbuilder functionality, including entry manipulation, structured data handling, buffer operations, and memory failure scenarios
2. Fixed memory allocation check in stumpless_get_entry_hostname() and stumpless_get_entry_procid() to properly handle allocation failures
3. Added build configuration in tools/cmake/function_tests.cmake

Motivation
The strbuilder module was lacking sufficient test coverage. These tests help ensure the module's reliability and catch potential issues early.

Test Plan
All tests pass locally. Happy to make any adjustments if needed.


goatshriek (changes requested):
Because strbuilder is an internal utility, these tests should be placed with the portion of the code that they actually test, for example the entry tests for setting and getting hostname and procid should go into `test/function/etnry.cpp` as it operates on an entry.

The strbuilder error branches that were added should already be handled by the strbuilder error handling code for NULL values. Can you share a single tests that causes a failure without your changes to confirm they are necessary?

pengzhongChen:
## Update: Phase 1 Complete (Proof of Necessity), Phase 2 in Progress

Hi, I've made progress on the requested changes. Here is an update on the current status:

### Phase 1: Proof of Necessity
I have added regression tests in `test/function/entry.cpp` that trigger a **Segmentation fault** when the NULL checks in `src/entry.c` are removed. This confirms that the fix is necessary for memory safety.

**Test 1 - GetHostName.MallocFailureOnStrbuilder**
```cpp
TEST( GetHostName, MallocFailureOnStrbuilder ) {
  void * (*set_malloc_result)(size_t);
  const char *hostname;
  struct stumpless_entry *entry;

  entry = create_entry(  );
  ASSERT_NOT_NULL( entry );

  set_malloc_result = stumpless_set_malloc( MALLOC_FAIL );
  ASSERT_NOT_NULL( set_malloc_result );

  hostname = stumpless_get_entry_hostname( entry );
  EXPECT_NULL( hostname );
  EXPECT_ERROR_ID_EQ( STUMPLESS_MEMORY_ALLOCATION_FAILURE );

  set_malloc_result = stumpless_set_malloc( malloc );
  EXPECT_TRUE( set_malloc_result == malloc );

  stumpless_destroy_entry_and_contents( entry );
  stumpless_free_all(  );
}
```
**Test 2 - GetProcid.MallocFailureOnStrbuilder**
```cpp
TEST( GetProcid, MallocFailureOnStrbuilder ) {
    void * (*set_malloc_result)(size_t);
    const char *procid;
    struct stumpless_entry *entry;

    entry = create_entry(  );
    ASSERT_NOT_NULL( entry );

    set_malloc_result = stumpless_set_malloc( MALLOC_FAIL );
    ASSERT_NOT_NULL( set_malloc_result );

    procid = stumpless_get_entry_procid( entry );
    EXPECT_NULL( procid );
    EXPECT_ERROR_ID_EQ( STUMPLESS_MEMORY_ALLOCATION_FAILURE );

    set_malloc_result = stumpless_set_malloc( malloc );
    EXPECT_TRUE( set_malloc_result == malloc );

    stumpless_destroy_entry_and_contents( entry );
    stumpless_free_all(  );
}
```
Verification Results:

- **Without the fix**: Tests consistently crash with a segmentation fault.
- **With the fix applied**: All tests pass successfully.

### Phase 2: Refactoring Tests
Following your suggestion, I am currently moving the relevant tests from test/function/strbuilder.cpp to test/function/entry.cpp. This will consolidate entry-related functional tests into the appropriate suite.

I will push the updates once Phase 2 is finalized. Please let me know if you have any feedback on this approach so far!

pengzhongChen:
Done as requested. Here's a summary of the changes:

* **Added** `GetProcid.MallocFailureOnStrbuilder` and `GetHostName.MallocFailureOnStrbuilder` tests to demonstrate that the `strbuilder_new()` NULL check is needed.
* **Moved** `EntryToBuffer` tests from `strbuilder.cpp` to `entry.cpp` since they test entry-to-buffer functionality.
* **Removed** the `strbuilder.cpp` test file.

Please let me know if further adjustments are needed. Thanks for the review!

goatshriek (changes requested):
Thank you for clarifying the precise tests that cause the failures, this is helpful. The strbuilder interface is defined to work by propagating failures through the calls to avoid repeated checks in all callers, and it would be more consistent with this interface to do this NULL check in `strbuilder_destroy` where the crash occurs, instead of multiple callers. Does this still resolve your failing tests?

I haven't gone through these new tests in detail, but before doing that I want to ask if you've made sure these aren't duplicative of existing tests? You are adding quite a few new ones.

pengzhongChen:
Thank you for the suggestion! I've refactored the code as you recommended:

1. **Added NULL check in `strbuilder_destroy()`** (`src/strbuilder.c`):
   - Added `if (!builder) return;` before `cache_free()` to prevent crashes when `strbuilder_destroy(NULL)` is called

2. **Removed the NULL checks from `entry.c`**:
   - Removed the `if (!hostname_builder)` and `if (!procid_builder)` checks from `GetHostName` and `GetProcid`

This approach is cleaner and handles the NULL case uniformly for all callers of `strbuilder_destroy()`.

3. **Reviewed the new tests**:
   - After reviewing the existing test suite, I removed the 8 duplicate tests I had added
   - Kept 2 non-duplicate tests that provide additional coverage:
     - `EntryToBufferTest.MultiByteCharacters` - verifies multi-byte UTF-8 characters (Chinese and Russian) are correctly written to buffer
     - `EntryToBufferTest.AppendStructuredDataEmpty` - verifies syslog format when structured data is empty

These changes are now passing all local tests. Let me know if there’s anything else!

pengzhongChen:
Hi, I noticed the overall Codecov coverage dropped slightly in this PR, and I wanted to check in gently.

All of my changed lines are still 100% covered by tests, but I’m happy to add more test cases or adjust anything else if that would help bring the coverage back up or improve the implementation further.

The PR is ready whenever you have a moment to review, and please let me know if there’s anything I can refine or fix on my end!

test/function/entry.cpp
```diff
@@ -3022,7 +3066,7 @@ namespace {
     stumpless_destroy_entry_and_contents(large_entry);
   }
 
-TEST( EntryDestructTest, FreeUncachedEntry ) {
+  TEST( EntryDestructTest, FreeUncachedEntry ) {
```
goatshriek: This indentation got inadvertently changed, please switch it back.

test/function/entry.cpp
```diff
@@ -3034,4 +3078,67 @@ TEST( EntryDestructTest, FreeUncachedEntry ) {
   stumpless_destroy_entry_only(&manual);
 }
 
+  TEST( EntryToBufferTest, AppendStructuredDataEmpty ) {
```
goatshriek: This test is duplicative of some tests in test/function/target/buffer.cpp, specifically `BufferTargetTest, Basic` and `BufferTargetTest, EmptyMessage`. A better approach would be to enhance those with a more targeted check for the structured data element of the formatted result.

test/function/entry.cpp
```diff
@@ -3034,4 +3078,67 @@ TEST( EntryDestructTest, FreeUncachedEntry ) {
...
+    EXPECT_NO_ERROR;
+
+    EXPECT_THAT( buffer, testing::HasSubstr( "- " ) );
+
+    stumpless_destroy_entry_and_contents( entry );
+    stumpless_close_buffer_target( target );
+    stumpless_free_all(  );
+  }
+
+  TEST( EntryToBufferTest, MultiByteCharacters ) {
```
goatshriek: this test should also go into test/function/target/buffer.cpp rather than the entry test suite.

goatshriek (changes requested):
Thanks for sticking with this through the requested changes. Some of the new tests need to be reviewed for duplication, and are more appropriate in the buffer target test suite rather than the entry one, since they deal with how the buffer target formats and writes the message rather than how the entry is handled. Once this are cleaned up this should be good to go.

goatshriek:
Ah sorry, I forgot to answer your question about coverage. The short answer is that it isn't something you've done, and you don't need to worry about it, but thank you for being mindful of it!

There is a small bit of variability in the test suite coverage, specifically around which code paths are exercised for less deterministic code paths like the networking code. You don't need to worry about reported small swings in the number, just make sure that your patch/intended code is now covered, which you have done.

pengzhongChen:
Thanks for the review! I've addressed all the feedback:

- Fixed the indentation issue in `EntryDestructTest, FreeUncachedEntry`
- Moved `EntryToBufferTest, AppendStructuredDataEmpty` and `EntryToBufferTest, MultiByteCharacters` from `entry.cpp` to `test/function/target/buffer.cpp`
- Enhanced `BufferTargetTest, EmptyMessage` with a targeted check for the structured data element (`-`)
- Added `BufferTargetTest, MultiByteCharacters` for multi-byte character coverage

Thanks for explaining the coverage variability — that clears things up!
All tests are passing. Let me know if there are any other issues!

goatshriek (approved):

PR #565, opened 2026-04-19
Add failure and boundary coverage for param tests

chithien14td-beep:
## Summary

This change adds missing failure-path and boundary coverage for public param APIs.

## What changed

* Added malloc failure coverage for `stumpless_copy_param`
* Added deeper allocation failure coverage for `stumpless_copy_param`
* Added boundary coverage for `stumpless_param_into_string` with:

  * zero-size/null buffer handling
  * minimal non-zero buffer handling
  * near-threshold insufficient buffer handling

## Notes

These tests stay within the existing `test/function/param.cpp` structure and follow the current function test style. The `param` test target passes locally with these additions.

### DIFF
--- a/test/function/param.cpp
+++ b/test/function/param.cpp
@@ -230,6 +230,68 @@ namespace {
     free( buffer );
   }
 
+  TEST_F( ParamTest, ParamIntoStringBufferSizeZero ) {
+    size_t result;
+
+    result = stumpless_param_into_string( &basic_param, NULL, 0 );
+    EXPECT_EQ( result, 0 );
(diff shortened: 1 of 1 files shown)


chithien14td-beep:
Thanks for taking a look. I’ve cleaned up this PR so it only contains the param test coverage changes. The new tests pass locally, and the checks are now passing on GitHub as well. Please let me know if there’s anything else I should adjust.

goatshriek:
I only see one commit and no mention of a push after I looked at this before; do you need to push the clean up that you mentioned?

PR #573, opened 2026-06-02
docs: add Simplified Chinese basic usage guide

Lin-ZJ:
Adds a Simplified Chinese translation of the basic usage guide.

Also renames the localized example directory from `简单` to `基础`, which reads more naturally in Chinese.

Verification:
- Ran `git diff --check upstream/latest...HEAD`
- Verified the relative Markdown link

Refs #117.


goatshriek:
Thanks very much for putting this together! Translation contributions are always greatly appreciated, especially when they can correct issues in existing translations.

